PR #46, opened 2020-11-30
et.json

Repsionu:
et-EE translation for 1.3.2


Repsionu:
So rarely use git nowdays... No idea how to refresh/sync my fork with master :-(

Repsionu:
Could not get verified :-(

nielsfaber:
I don't have a problem with that. Your PR cannot really break anything.
The only problem i have with the translations is to check if they are complete (and remain complete).
As far as i know, there is not really a github check that could do something like that...

PR #50, opened 2020-12-03
Some recent translation changes

Repsionu:
Added Estonian (et-EE)
Best, JR


nielsfaber:
Thanks! i will try to make a release with it soon

PR #115, opened 2021-01-28
Create fr.json

Patrick44fr:
hi, it's my first contribution, but i don't understand github very well


nielsfaber:
I assume the other one is more recent

PR #142, opened 2021-02-19
Removed unused code

ludeeus:
Was looking at the code and it seems that this is no longer in use, this PR clean up the unused code.


nielsfaber:
Thanks for this! 👍

PR #167, opened 2021-03-31
Minor changes for consistency

shred86:
Some very minor changes I noted as I was reading through the documentation and setting up Alarmo.

I also noticed in the documentation it states:

"Note that assigning a sensor type is not mandatory, and all configuration settings can also be set manually. It is also possible to deviate from the predefined configuration after setting a type."

But when I set a Device Type in the Alarmo panel, certain sensors only display certain predefined configurations (i.e. Advanced settings). I'd be happy to update this in the documentation to reflect that depending on the device type, you will only see certain settings but I wasn't sure if this was by design or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ This is an alarm system integration for Home Assistant. It provides a user inter
     - [Sensor configuration](#sensor-configuration)
       - [Sensor types](#sensor-types)
       - [Immediate](#immediate)
-      - [Always-on](#always-on)
-      - [Allow-open](#allow-open)
-      - [Arm on close](#arm-on-close)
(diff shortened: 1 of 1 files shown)


nielsfaber:
Thanks for caring, documentation is not my strongest point 😅 

As for:
>  I'd be happy to update this in the documentation to reflect that depending on the device type, you will only see certain settings but I wasn't sure if this was by design or not.

These settings and their default values are all made to help you with the setup.
In fact, setting a device type _only_ results in the settings being filtered (to the ones that are useful) and having some initial settings applied, so this is very much by design.
If it's unclear or you miss options of course this is open for discussion.

PR #201, opened 2021-05-09
Add spanish translation

ajtudela:
I added spanish translations to localization folder. 

Keep doing such a great work!


nielsfaber:
Thanks!
I will make sure to include it in the next release.

PR #216, opened 2021-05-31
Only allow once instance of the integration to be installed

rroller:
Without this change, the user can attempt to install two instances of alarm, which will put the system into a bad state (which I did by a mis click).

NOTE: I haven't tested this, but I believe this is the standard way to make only 1 instance installable, see https://github.com/home-assistant/core/search?q=_async_current_entries


nielsfaber:
This looks really promising!
I was looking into this a while ago but couldn't find a way to hide the option in GUI (also not for the 'reloading' of the integration, which causes the configuration to be wiped).
Catching the case during setup didn't cross my mind, but it seems a good alternative.
I will run some tests for this, hopefully I can also make an extension to block the reloading as well.
Thanks!

PR #227, opened 2021-06-09
Update nl.json

Tosko4:
Few english strings remained, translated that and one string changed it slightly a bit to make it better readable


nielsfaber:
goed gezien!

PR #229, opened 2021-06-10
Update fr.json

Patrick44fr:
tanks for alarmo work.


nielsfaber:
Thanks for keeping maintaining the translation!

PR #257, opened 2021-08-03
Localization

mattexpa:
added de.json file fully translated

### DIFF
--- a/custom_components/alarmo/frontend/localize/languages/de.json
+++ b/custom_components/alarmo/frontend/localize/languages/de.json
@@ -1,0 +1,494 @@
+{
+	"common": {
+		"modes_long": {
+			"armed_away": "Scharf - Unterwegs",
+			"armed_custom_bypass": "Scharf - Benutzerdefiniert",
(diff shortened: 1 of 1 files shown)


ConsoleCriminal (approved):

TheZoker (approved):

nielsfaber:
Vielen Dank! 👍 
I will make sure its in the next update.

nielsfaber:
@mattexpa
On second thoughts: your translation file seems completely scrambled, it doesn't have the same structure as the translations.
Please compare:
https://github.com/nielsfaber/alarmo/blob/main/custom_components/alarmo/frontend/localize/languages/en.json
with
https://github.com/nielsfaber/alarmo/blob/main/custom_components/alarmo/frontend/localize/languages/de.json

Seems like all JSON entries are alphabetized, from line 29 it is too much mixed up to recognize.
I cannot maintain the file this way since I cannot compare across files for missing entries.

I don't know how to correct this myself.
Could you make a new PR with the translation file in the correct structure?
You can commit to main branch.
Thanks!

PR #343, opened 2021-12-03
Update fr.json

Patrick44fr:
A new version with wilcard

i hope is good


nielsfaber:
Yes looks good, thank you for fixing it!

PR #429, opened 2022-03-23
Update sv.json

Rittsel:
Added translation for vacation mode, corrected spelling mistakes and rewrote some translations.


nielsfaber:
Thanks a lot for providing this 👍 
Hope you can also check back every once in a while, if things change in Alarmo some translations may be affected.

PR #436, opened 2022-04-03
Fixed Translation SyntaxError: MALFORMED_ARGUMENT in Spanish language

kikorr:
The syntax was wrong and shows the Translation SyntaxError: MALFORMED_ARGUMENT error making the Alarmo panel don't work as expected in Spanish language.


nielsfaber:
Thanks, I would probably never have found this myself 😅

PR #533, opened 2022-07-25
Treat CO sensors as Environmental

berg:
Hi there! This PR treats binary sensors of type `carbon_monoxide` as Environmental sensors.

I'm afraid this PR is somewhat incomplete as I couldn't get the build/compile steps to run locally without creating a huge diff, so I didn't update `dist/alarm-panel.js`. Apologies if I missed any docs on how you'd like that done, but figured the PR would be easier to read without a bunch of unrelated changes!

Thanks so much for Alarmo—I'm just getting started with it but it seems really useful!


nielsfaber:
Thanks a lot 👍 
The dist/alarm-panel.js is not relevant for a PR. It will be compiled from the source files before making a new release.

PR #580, opened 2022-09-23
Update sv.json

newlund:
Fix errors and typos


newlund:
Should fix this

image

nielsfaber:
Ooff I didn't know of this error, it looks very bad. Thanks for fixing it!

PR #619, opened 2022-12-09
Update it.json

g1za:
Translated "Add".
Fixed a capital letter (even if not relevant as the interface shows everything upper case).


g1za:
Another question: Alarmo card shows “Disattivo” when the alarm is deactivated. Actually “Disattivo” means “deactivating” and it would be ok while the alarm is being deactivated. When this is done it should show “Disattivato” (Deactivated). I looked for this as I wanted to improve the translation but I could not find it in the code. Where should I look for it?

nielsfaber:
Sorry for the delay!

PR #690, opened 2023-04-05
Update hacs.json

dennisstorm-lab:
Set the HASS version to the most recent non-BETA version

### DIFF
--- a/hacs.json
+++ b/hacs.json
@@ -4,5 +4,5 @@
   "zip_release": true,
   "filename": "alarmo.zip",
   "hide_default_branch": true,
-  "homeassistant": "2023.4.0b1"
+  "homeassistant": "2023.3.6"
 }


nielsfaber:
Why is this needed? I understand from HACS documentation that the non-beta releases will also be passing this condition.

dennisstorm-lab:
You have the following in hacs.json: homeassistant: "2023.4.0b1"

The docs state: "Allow Home Assistant beta versions by appending b0. Without b0, only official releases will be allowed."
This is ofcourse perfectly valid, since you only want to deploy on official HASS releases.

However, since 2023.4.0b1 is HIGHER than 2023.3.6 nothing gets installed, which is the issue.

The 'homeassistant" value should contain the minimum required version. Since practically no-one has a BETA running, which is higher than the current HASS release installation will not happen.

image

nielsfaber:
My intention was to have v1.9.7 show up as latest available version for users who are on HA 2023.3 or lower.
And to have v1.9.8 show up for users who are on HA 2023.4 (beta or alpha).
If I understand you correctly, pinning to HA 2023.4 in the hacs.json file causes the entire Alarmo project to require HA 2023.4, which is not what I wanted.
In this case I rather remove the pinning completely, HACS documentation is not very clear on this.

nielsfaber:
I tried this in my own HA installation.
With HA 2023.3 installed.

V1.9.7 shows up as OK for download:
image

V1.9.8 shows up as incompatible:
image

After updating to HA 2023.4, V1.9.8 shows up as available for download:
image

All seems to work as intended so we can close this PR.

PR #727, opened 2023-05-14
Update package versions and fix minor lint issue

oded996:
Fix versioning and lint issues to allow compliation of alarmo frontend


custom_components/alarmo/frontend/package.json
```diff
@@ -30,17 +30,17 @@
     "@typescript-eslint/eslint-plugin": "^2.6.0",
     "@typescript-eslint/parser": "^2.6.0",
     "ajv": "^6.12.6",
-    "custom-card-helpers": "^1.6.5",
+    "custom-card-helpers": "1.6.5",
```
nielsfaber, line 33: I don't want to pin to fixed package versions. 
We should use latest-greatest whenever possible.

custom_components/alarmo/frontend/package.json
```diff
@@ -30,17 +30,17 @@
...
     "eslint": "^6.8.0",
     "eslint-config-airbnb-base": "^14.0.0",
     "eslint-config-prettier": "^6.5.0",
     "eslint-plugin-import": "^2.18.2",
     "eslint-plugin-prettier": "^3.1.1",
-    "home-assistant-js-websocket": "^5.7.0",
+    "home-assistant-js-websocket": "5.7.0",
```
nielsfaber, line 39: I don't want to pin to fixed package versions. 
We should use latest-greatest whenever possible.

nielsfaber (changes requested):
Thanks for the PR!
I have some comments in the changes you made to the packages list.

I was wondering what linting configuration you are using.
The changes look OK to me, but I would expect issues in almost every file based on the corrections you have made.
Did you only lint these specific files for a reason?

custom_components/alarmo/frontend/package.json
oded996: There seems to be a conflict between custom-card-helpers and home-assistant-js-websocket
See logs below. 
I am not sure how to resolve it. Do you have any suggestions?


> alarmo@1.0.0 rollup
> rollup -c


src/alarm-panel.ts → dist...
[!] (plugin rpt2) Error: /Users/odeds/Alarmo/alarmo/custom_components/alarmo/frontend/src/helpers.ts(14,20): semantic error TS2345: Argument of type 'import("/Users/odeds/Alarmo/alarmo/custom_components/alarmo/frontend/node_modules/home-assistant-js-websocket/dist/types").HassEntity' is not assignable to parameter of type 'import("/Users/odeds/Alarmo/alarmo/custom_components/alarmo/frontend/node_modules/custom-card-helpers/node_modules/home-assistant-js-websocket/dist/types").HassEntity'.
  Type 'HassEntity' is not assignable to type 'HassEntityBase'.
    Types of property 'context' are incompatible.
      Property 'parent_id' is missing in type '{ id: string; user_id: string | null; }' but required in type 'Context'.
src/helpers.ts
Error: /Users/odeds/Alarmo/alarmo/custom_components/alarmo/frontend/src/helpers.ts(14,20): semantic error TS2345: Argument of type 'import("/Users/odeds/Alarmo/alarmo/custom_components/alarmo/frontend/node_modules/home-assistant-js-websocket/dist/types").HassEntity' is not assignable to parameter of type 'import("/Users/odeds/Alarmo/alarmo/custom_components/alarmo/frontend/node_modules/custom-card-helpers/node_modules/home-assistant-js-websocket/dist/types").HassEntity'.
  Type 'HassEntity' is not assignable to type 'HassEntityBase'.
    Types of property 'context' are incompatible.
      Property 'parent_id' is missing in type '{ id: string; user_id: string | null; }' but required in type 'Context'.

custom_components/alarmo/frontend/package.json
oded996: This one is actually not nessesary. Reverted.

oded996:
> Fix versioning and lint issues to allow compliation of alarmo frontend

I did not create any linting configuration. 
Clone the repository, then run npm run build

All of the lint fixes are completely automatic except for alarmo-select.ts which used objects before they defined, so the file order had to be changed.

nielsfaber:
> There seems to be a conflict between custom-card-helpers and home-assistant-js-websocket
See logs below.
I am not sure how to resolve it. Do you have any suggestions?

I am not aware of the conflict, but I believe you.
It [appears](https://github.com/search?q=repo%3Anielsfaber%2Falarmo%20custom-card-helpers&type=code) mostly the `HomeAssistant` and `fireEvent` definitions are used.
I guess we can get rid of custom-card-helpers completely.
It is very poorly maintained / outdated anyway.

My proposal would be to create a local copy of the functions which are used.
Their license allows copying their code.

I am able to make this change, but it might take a week or 2 before I have the time to do so.

PR #747, opened 2023-06-13
Allow to set "bypass_open_sensors" and "skip_delay" when arm the alarm using a mqtt command

robertoprubio:
Allow to include "bypass_open_sensors" (or "force") modifier as a MQTT command parameter to allow the alarm to be armed event if open sensors.
Also allow to include "skip_delay" modifier as a MQTT command parameter to allow the alarm to be armed immediately.


nielsfaber:
thanks for your contribution, looks good to me 👍

PR #832, opened 2023-11-02
Add Vietnamese translations

vinhtantran:
Add Vietnamese translated messages and options to localization


nielsfaber:
It looks good, thank you for your work!

Issue #827, opened 2023-11-01
Add `delay` wildcard option


mineiwik:
### Checklist
- [x] I checked for similar existing requests (both open and closed) before posting.
- [x] My request is generic, other users may benefit from it too.

### Proposal
Add another wildcard option to allow the `delay` attribute for `arming` and `pending` states to be added to the notification action message.

### Additional info
The benefit here is that the same notification message template can be used for all modes, which may have different delay values.
I utilize the delay value in the TTS notification to communicate, for instance, the remaining time until the alarm goes off.
I would be willing to implement this myself.

PR #834, opened 2023-11-04
Add `delay` wildcard option

mineiwik:
Implements feature request from issue #827


nielsfaber:
This looks very good, I would have made it in the same way 👍

PR #848, opened 2023-11-28
Register as config panel

jpbede:
Register alarmo panel as config panel like ZHA/Matter/... instead of in the sidebar. This capability was introduced mid this year (https://github.com/home-assistant/core/pull/96245) in the HA core.

I don't know if we should keep the entry in the sidebar for reasons of "backwards compatibility" so as not to confuse users.

https://github.com/nielsfaber/alarmo/assets/3989428/86b73838-6dc4-4b9b-af1c-9313e014bdbb

### DIFF
--- a/custom_components/alarmo/panel.py
+++ b/custom_components/alarmo/panel.py
@@ -35,10 +35,9 @@ async def async_register_panel(hass):
         webcomponent_name=PANEL_NAME,
         frontend_url_path=DOMAIN,
         module_url=PANEL_URL,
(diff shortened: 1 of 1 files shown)


nielsfaber:
Hi, thanks for your contribution!
Personally (and I think this also applies to other users), I prefer to keep Alarmo in the sidebar. It's much easier and quicker to find it back.
If you could change your PR to maintain this compatibility, i'm fine with adding this.

jpbede:
Thanks, re-added the sidebar icon

PR #916, opened 2024-03-31
Avoid use of deprecated `hass.helpers`

jpbede:
With the HA release 2024.5 we start deprecating the use of `hass.helpers` (https://developers.home-assistant.io/blog/2024/03/30/deprecate-hass-helpers/). This will be a breaking change with version 2024.11.


nielsfaber:
thanks for your contribution 👍

Issue #952, opened 2024-05-30
Unable to disarm Alarmo on Home Assistant 2024.6.0 beta


krazos:
### Alarmo Version

v1.10.1

### HA Version

2024.6.0b1

### Bug description

After upgrading to core-2024.6.0b0, I am no longer able to disarm Alarmo directly or through a script or automation. See log entries below.

### Steps to reproduce

Triggered automation/ran script that successfully disarmed Alarmo prior to core-2024.6.0b0. Also tried disarming Alarmo manually with code. 

### Relevant log output

```shell
Logger: homeassistant.components.automation.disarm_alarmo_********
Source: helpers/script.py:526
integration: Automation (documentation, issues)
First occurred: 5:30:00 AM (1 occurrences)
Last logged: 5:30:00 AM

Disarm Alarmo ********: Error executing script. Unexpected error for call_service at pos 1: AlarmoBaseEntity.async_alarm_disarm() takes 1 positional argument but 2 were given
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 526, in _async_step
    await getattr(self, handler)()
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 763, in _async_call_service_step
    response_data = await self._async_run_long_action(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 726, in _async_run_long_action
    return await long_task
           ^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2741, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2784, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 969, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1041, in _handle_entity_call
    result = await task
             ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/alarm_control_panel/__init__.py", line 207, in async_handle_alarm_disarm
    await self.async_alarm_disarm(self.code_or_default_code(code))
[...]

PR #953, opened 2024-05-30
Make code a positional argument for async_alarm_disarm

jpbede:
With HA 2024.6.0 `code` needs to be a positional argument to `async_alarm_disarm`.

Fixes https://github.com/nielsfaber/alarmo/issues/952

### DIFF
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -307,16 +307,18 @@ class AlarmoBaseEntity(AlarmControlPanelEntity, RestoreEntity):
             self._changed_by = res[ATTR_NAME]
             return (True, res)
 
-    async def async_service_disarm_handler(self, code, context_id=None):
+    @callback
+    def async_service_disarm_handler(self, code, context_id=None):
         """handle external disarm request from alarmo.disarm service"""
         _LOGGER.debug("Service alarmo.disarm was called")
 
-        await self.async_alarm_disarm(
+        self.async_alarm_disarm(
             code=code,
             context_id=context_id
         )
 
-    async def async_alarm_disarm(self, code, **kwargs):
+    @callback
+    def alarm_disarm(self, code, **kwargs):
         """Send disarm command."""
         _LOGGER.debug("alarm_disarm")
         skip_code = kwargs.get("skip_code", False)
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -307,21 +307,18 @@ class AlarmoBaseEntity(AlarmControlPanelEntity, RestoreEntity):
             self._changed_by = res[ATTR_NAME]
             return (True, res)
 
-    @callback
-    def async_service_disarm_handler(self, code, context_id=None):
+    async def async_service_disarm_handler(self, code, context_id=None):
         """handle external disarm request from alarmo.disarm service"""
         _LOGGER.debug("Service alarmo.disarm was called")
 
-        self.async_alarm_disarm(
+        await self.async_alarm_disarm(
             code=code,
             context_id=context_id
         )
 
-    @callback
-    def async_alarm_disarm(self, **kwargs):
+    async def async_alarm_disarm(self, code, **kwargs):
         """Send disarm command."""
         _LOGGER.debug("alarm_disarm")
-        code = kwargs.get("code", None)
         skip_code = kwargs.get("skip_code", False)
         context_id = kwargs.get("context_id", None)
 


custom_components/alarmo/alarm_control_panel.py
```diff
@@ -307,21 +307,18 @@ def _validate_code(self, code, to_state):
             self._changed_by = res[ATTR_NAME]
             return (True, res)
 
-    @callback
-    def async_service_disarm_handler(self, code, context_id=None):
+    async def async_service_disarm_handler(self, code, context_id=None):
```
gjohansson-ST, line 310: I think it's intentional these was not async before so maybe something else breaks?

jpbede: Removed the async_ prefix then, otherwise we await this

gjohansson-ST: Well, this integration isn't exactly written to standard so no idea. Was just a recommendation to maybe not introduce a new issue

nielsfaber: @jpbede Thanks a lot for your initiative once again 👍 

@gjohansson-ST Could you explain why the argument `code` was added as positional argument only for service `async_alarm_disarm`? Shouldn't this be the same for `async_alarm_arm_away` etc. ?

 I read your blog post about this but it did not mention this breaking change, this post only tells developers to implement `code_arm_required` (at least thats what I got out of it). Perhaps you could describe the complete change?

The developer docs also need a small update, the code examples still show `code` as keyword argument, see: https://github.com/home-assistant/developers.home-assistant/blob/master/docs/core/entity/alarm-control-panel.md

EvertJob:
When will the update be pushed?

custom_components/alarmo/alarm_control_panel.py
gjohansson-ST: @nielsfaber I have no idea what you're referring too. All the methods are the same and has been for years.
Also the examples in the dev docs are correct and matching what is in the code.

Your methods signatures however is different than what should be implemented but until now there was no validation (but now there is) hence why we get an issue here. If you need them different I suppose you should make other methods than trying to override the default ones.

nielsfaber: @gjohansson-ST 
I will try to make myself more clear by breaking it down in separate points:

1. As @jpbede stated as background for the PR: _"With HA 2024.6.0 code needs to be a positional argument to async_alarm_disarm."_ On the contrary you say: _"All the methods are the same and has been for years."_. These are conflicting statements, I believe a change is needed but for me it comes as a surprise.
2. Why is the change (that is discussed in this PR) only about the method `async_alarm_disarm`? Since next to disarming also arming methods may require code input, I would expect methods `async_alarm_arm_away`, `async_alarm_arm_home`, etc. to be affected as well (i.e. needing extra positional argument for `code`) for sake of consistency.
3. Why is breaking change not described in your blog post [here](https://developers.home-assistant.io/blog/2024/05/22/alarm_control_panel_validation)? I would say it is quite important to notify custom component developers that this change is upcoming. I also would like to know the reason behind the change.
4. The code examples in the [developer docs](https://github.com/home-assistant/developers.home-assistant/blob/master/docs/core/entity/alarm-control-panel.md) still use `code` as keyword argument rather than positional argument. I suggest to update the document to give a good reference/example to developers.

gjohansson-ST: 1. You have to ask him for the subject of the PR. With 2024.6 there is a validation implemented (with the use of setting a default code). None of which are changing the methods signatures so in essence the PR subject line is wrong but I guess he wrote it like that as you have been overwriting the methods incorrectly.
2. Don't know. The change in `alarm_control_panel` are for all methods
3. There is nothing breaking if you follow what's been in the dev docs for years (which is not the case as you use different inputs to the methods). So from our point of view there isn't anything breaking and as such not noted on the dev blogs. It's impossible for us to handle if custom component devs are implementing the methods incorrectly.
4. As earlier mentioned. The dev docs are correct. `code` is the only parameter and should have a default value of `None`. No other parameters are allowed as that simply doesn't work with the validation (hence why you need to use a custom service if you need something else).

jpbede: 1. I'm sorry if the PR description/title was misleading that it was a method signature change within HA.
2. TBH I haven't checked the other methods of alarmo, I just fixed this I saw this on my own system and saw it got mentioned on Discord.

nielsfaber: It’s still very strange to me that this PR is needed to restore compatibility with HA 2024.6 if nothing was changed on the side of HA. But given that this fix is proven to be effective, I’m merging this and will make a new release of alarmo for it.

EvertJob: > It’s still very strange to me that this PR is needed to restore compatibility with HA 2024.6 if nothing was changed on the side of HA. But given that this fix is proven to be effective, I’m merging this and will make a new release of alarmo for it.

I don’t have the knowledge of ha in any advanced way but I found out that you can turn off the alarm under developer tools -> services.

SEKCobra:
I just looked at the changelog of 2024.6, but I can't find this breaking change listed anywhere. How did you find out about this change? I just had to handle a police visit due to the inability to disable the triggered alarm in time and want to improve my update process in the future.
Admittedly I had missed the Alarmo update, but I am always diligently studying the update notes for breaking changes and still can't find anything about this anywhere.

gjohansson-ST:
> I just looked at the changelog of 2024.6, but I can't find this breaking change listed anywhere. 

Because it wasn't a breaking change. Alarmo implements the methods incorrectly so therefore if something is added in the `ClimateEntity` which doesn't fit Alarmo's way of working it's breaking.
I would recommend it to be reworked but that's up to someone to fix.

PR #1115, opened 2025-02-13
Added Catalan translation

fenikso:
Added ca.json file for backend translation.


nielsfaber:
Thank you for providing this translation. I will include this in the next update.

PR #1138, opened 2025-04-11
Documentation Update: Clarified the way to set up a split configuration

gregorylie:
Clarified how to configuration both a UI based and configuration.yaml bridge configuration to support the pin in Homekit


nielsfaber:
Thanks for your contribution! 👍

PR #1158, opened 2025-05-12
add sensor config services

cayossarian:
Adds the ability to modify Alarmo sensor properties including turning the Alarmo sensor on or off.  These new actions are now available to automations that need to dynamically change properties or temporarily disable or enable an Alarmo sensor while keeping the underlying HA sensor active or otherwise adjust the properties of the Alarmo sensor.

Service data validation is performed before any changes are allowed.  If properties are missing from the service call the properties of the storage are used, allowing one to pass only the properties that need to be modified.  The exception is that the boolean for bypass, if true, must be accompanied by the bypass modes.  Bypass modes are verified against the enabled modes and if they do not align the service call is rejected with an appropriate error.

What this change does not do is update the frontend without a manual refresh of the browser.  For a refresh to automatically occur the frontend needs to listen for a change event which is indeed set by the service backend.

Tests are provided in the test directory (see the comment at the top of the test file).

Developer Note:  The Alarmo repository seems to use CRLF in its files which is not typical and this causes problems for users trying to commit directly from Unix style systems.  This issue might be avoided by adding a .gitattribute file to the repository - see [https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/](https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -592,6 +592,64 @@ Structure of the event data:
 }
 ```
 
+---
+
+Alarmo supports two admin services for advanced sensor configuration:
+
+**Disable an Alarmo sensor**
+
+Example YAML:
+```yaml
+service: alarmo.set_sensor_configuration
+data:
+  entity_id: binary_sensor.virtual_window_sensor_window
+  enabled: false
+```
+
+**`alarmo.set_sensor_configuration`**  
+Update Alarmo-specific configuration for one or more sensors. Supports updating type, enabled state, modes, and advanced flags (e.g., auto_bypass).
+
+Example YAML:
+```yaml
+service: alarmo.set_sensor_configuration
+data:
+  entity_id:
+    - binary_sensor.virtual_motion_sensor_motion
+    - binary_sensor.virtual_window_sensor_window
+  type: door
+  modes:
+    - armed_away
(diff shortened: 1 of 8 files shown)


nielsfaber:
@cayossarian 
Thank you for taking initiative for improving alarmo.
Unfortunately I don't think we have had any discussion/alignment about this feature before you started implementing this.
Frankly I don't see a need for the sensor settings to be changed via automation.
It seems we are not aligned on this, please reach out (open a feature request) to discuss it before submitting any PR for review.

I see your PR also includes a unit test, which is appreciated, but since the rest of the project does not include any tests, the total coverage remains close to zero.
Unfortunately this is beyond of my skill level, but if you wish to help out with this, perhaps it makes sense to start setting up tests for the entire code base.

cayossarian:
No worries @nielsfaber, I assumed the utility would be self-evident in being usable and I tried to ensure no logic in the existing system was not affected but it's your repository so no worries.  I'll close this request and use my custom repository.  Cheers.

PR #1166, opened 2025-05-22
Adds Turkish translations

mcanvar:
Thank you for this handy extension. I've translated each keyword. If you accept, I will check on place if there is any chicken translation.


nielsfaber:
Thank you very much for your efforts, I will include it in the next release.

PR #1170, opened 2025-05-31
fix: backend test fix for factory and expansion for ready to arm tests

cayossarian:
Fix an issue with the new required parameter for area config

Expand ready to arm tests

Update the test readme


nielsfaber:
Thank you very much again for your contribution!
The quality of the test and documentation seem excellent to me.

Issue #1177, opened 2025-06-05
Parallelize checking disarm codes to speed up disarming


therealmrfox:
### Checklist
- [x] I checked for similar existing requests (both open and closed) before posting.
- [x] My request is generic, other users may benefit from it too.

### Proposal
Calls to `alarmo.disarm` can take several seconds if there are multiple codes/users defined. I'm seeing ~6s for 13 users/codes. The [codes are checked in sequence](https://github.com/nielsfaber/alarmo/blob/64368ff62b8e7452f45e0fda414a5652a1403be4/custom_components/alarmo/__init__.py#L280) so the more users exist the longer this will take for new users. 

Ideally these checks should happen in parallel to make disarming responsive.

### Additional info
This is a rehash of https://github.com/nielsfaber/alarmo/issues/1070. Re-filing as a FR since I don't have the permission to re-open 1070 but the issue persists. I took a stab at sketching out the code to parallelize the checks in that bugreport but couldn't make it quite work.

Thank you!

therealmrfox:
Created a pull request (https://github.com/nielsfaber/alarmo/pull/1182) with a possible fix. Works for me on a RPi and has considerably decreased the time to disarm from 6s to 1s.

EVWorth:
@therealmrfox i know this was merged now, and i think it was a great addition after looking through your change. 

Question: is there a reason you hard coded the workers to 4 instead of letting concurrent futures auto decide the worker count based on cpu cores? By default it will assign workers equal to logical processors + 4

I wonder if it'd make sense to provide another PR updating this logic (just removing the hardcoded worker param) so that different systems with different processor counts can extract even more performance.

Maybe you tested already and you ran into issues. I'm curious to know!

therealmrfox:
@EVWorth having more threads than cores is useful for I/O-bound tasks, but here we're CPU bound, so it would only add more overhead. 

Most RPi devices seem to have 4 cores these days, so I just picked that. Even with more cores available, there will be diminishing returns unless someone has hundreds of users/codes. 

On more performant hardware this change is unlikely to matter much. When testing on my laptop I had to introduce artificial delays to see any change between old and new code. 


That said, perhaps instead of hard coding it, it could be `multiprocessing.cpu_count()`.

PR #1182, opened 2025-06-07
Use a thread pool to check user codes

therealmrfox:
Use multiple threads to check user codes to address slowness when disarming on slow hardware. This fixes https://github.com/nielsfaber/alarmo/issues/1177

### DIFF
--- a/custom_components/alarmo/__init__.py
+++ b/custom_components/alarmo/__init__.py
@@ -300,7 +300,7 @@ class AlarmoCoordinator(DataUpdateCoordinator):
             }
 
         with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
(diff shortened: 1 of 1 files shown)


therealmrfox:
I further tweaked the logic and decreased the number of rounds for bcrypt. This is a tradeoff between speed and security, but since an attacker would need access to HA or an unencrypted backup, that tradeoff is acceptable IMO. 

With the threadpool and reduced number of bcrypt rounds, disarming now takes .5s for me instead of 6s with 13 users. (The bcrypt change won't have an effect on existing users until they update their code).

nielsfaber:
Thanks a lot for this useful improvement!
I was aware of the computation power required for the encryption, but I didn't know there was a way to speed this up.

PR #1194, opened 2025-06-22
Update python codebase to replace use of .format() for f-strings in regular string format calls or %s in _LOGGER calls.

EVWorth:
I noticed that the codebase was using .format() for string formatting.

I refactored all _LOGGER calls to use %s string formatting, the logging module technically supports .format() and f-strings, but it's actually less efficient to use them. basically, you end up pre-parsing your string before passing the data to the logging module, rather than letting the logging module handle the concatenation under the hood (since it handles concatenation of variables using the %s syntax) 

I also refactored .format() calls not inside of _LOGGER calls to f-string since f-strings are easier to read, faster than .format() calls, and newer (introduced in Python 3.6)

References:
[https://docs.python.org/3/howto/logging.html#logging-variable-data](url)
[https://peps.python.org/pep-0498/](url)
[https://blog.finxter.com/string-formatting-vs-format-vs-formatted-string-literal/](url)

### DIFF
--- a/custom_components/alarmo/__init__.py
+++ b/custom_components/alarmo/__init__.py
@@ -341,7 +341,10 @@ class AlarmoCoordinator(DataUpdateCoordinator):
                 _LOGGER.info("Received request for disarming")
                 alarm_entity.alarm_disarm(None, skip_code=True)
             else:
-                _LOGGER.info("Received request for arming with mode {}".format(arm_mode))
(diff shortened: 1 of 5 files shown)


nielsfaber:
Thanks for letting me know, I was not aware of the difference.
I think I learned using the format() from reading HA's code, but I could be wrong or outdated.

Issue #1146, opened 2025-05-05
Avoid re-validation of sensors after triggered state ends


I-Gil:
### Checklist
- [x] I checked for similar existing requests (both open and closed) before posting.
- [x] My request is generic, other users may benefit from it too.

### Proposal
Currently, when the alarm is in an armed state and transitions to `triggered` due to a sensor activation, Alarmo attempts to re-arm the system after the `triggered` timeout expires.

This re-arming includes a fresh validation of the sensors, and if any are still open (e.g., a door remains open), the alarm fails to re-arm and becomes disarmed.

This is not ideal, as the system was already armed before entering `triggered`. It should return to the same armed state without re-validating the sensors.

I propose that Alarmo **restores the previous armed mode after the `triggered` state ends without running another sensor validation**.

### Additional info
This behavior would be more in line with traditional alarm systems, which remain armed after a trigger event unless manually disarmed.

It would help avoid unintentional disarming due to sensors that were already open during the intrusion and improve the reliability of the system during multi-event scenarios.

I-Gil:
Hi again, thanks for your reply and for maintaining this great integration.

What I gather from your comment is that the current behavior is intended as a workaround for the problem of how the system behaves during the triggered state:

When the alarm is triggered by a sensor, Alarmo sends a notification (this part works correctly).

However, if additional sensors are triggered while the alarm is already in the triggered state, no further notifications are sent — those activations are neither reported nor logged. (This is already covered in Issue #663, which I’ve also commented on).

As a result, I have no way to know whether other zones have been violated during the intrusion, which is a significant limitation.

To compensate for this lack of information, Alarmo attempts to automatically re-arm itself once the triggered period ends. But if any sensor is still open, the re-arming fails and the alarm is left disarmed — which (based on your comment) serves the purpose of forcing the user to check sensor states to verify what might have happened during the triggered period.

The issue is that the system ends up disarmed without the user's intention, simply because it couldn’t re-arm itself — this can go unnoticed and leave the property unprotected.

Proposal
If Alarmo sent a notification for every sensor activated during the triggered state, the need to check sensor states afterward would be significantly reduced — the user would already know what happened during the alarm.

Therefore:

I still believe the alarm should remain armed in its original mode after the triggered period ends, without attempting to re-arm.

Instead, the focus should be on reporting all triggered sensors during the triggered state, as already described in Issue #663.

This would keep the user well-informed and avoid unintentionally disarming the system when protection is still needed.

[...]

Issue #1103, opened 2025-01-20
Add attribute "last triggered alarm" to alarmo entity


Gifford47:
### Checklist
- [X] I checked for similar existing requests (both open and closed) before posting.
- [X] My request is generic, other users may benefit from it too.

### Proposal
I want to derive the last triggered alarm time by a template. It would be very easy to get this time (and date) from an attribute of the alarmo entity which is refreshed everytime the alarm triggers.
### Additional info
It's related to #642 and #960 

Please add an extra attribute with the last triggered alarm time as datetime type.

Ltek:
bump, very useful suggestion!

nielsfaber:
@Gifford47 Would something like this fit your needs?

![Image](https://github.com/user-attachments/assets/338542a5-62dd-4665-9716-f8f926d44288)

Gifford47:
Yes, that's great!

nielsfaber:
@Gifford47 OK great, I will make sure to include this in the next version update.
Thanks for your patience.

nielsfaber:
This feature is now released in v1.10.9.

PR #1202, opened 2025-06-26
Tests for last triggered attribute and re-arm after trigger expires with active sensors

cayossarian:
Covers feature requests
 #1103 Last Triggered Attribute
#1146 Avoid re-evaluation of sensors after trigger expires (rearm option)


nielsfaber:
Thanks a lot for expanding the tests 👍

Issue #1183, opened 2025-06-07
Identical codes between users should be disallowed


therealmrfox:
### Checklist

- [x] I am using the latest version of Alarmo (latest version can be found [here](https://github.com/nielsfaber/alarmo/releases/latest))
- [x] I checked for similar existing requests (both open and closed) before posting.

### Alarmo Version

v1.10.9

### HA Version

2025.5.3

### Bug description

Currently it is possible for 2 users to have the same arm/disarm code. 

This is a problem because users are identified by this code when they enter it. The following scenario is possible:

- Add user A to system (code: 1234, allow_disarm: True)
- Add user B to system (code: 1234, allow_disarm=False)

Now when the alarm is armed and user B enters their code, the alarm will disarm. (The opposite scenario is also possible where a user ends up being underprivileged). 

As users are looked up by their code, there really isn't a way to differentiate users A and B. So creation of user B with the same code should not be allowed. 

### Steps to reproduce

see above.

### Relevant log output

```shell

```

therealmrfox:
Fix: https://github.com/therealmrfox/alarmo/pull/1

nielsfaber:
I agree this is a useful addition.
Similarly to the code, the names of users should be unique, this is currently not verified by alarmo either.
I saw your PR has currently draft status, please submit it when ready.

therealmrfox:
Thanks @nielsfaber , good point about usernames also being unique. Added that to the PR. 

Ideally this should show error messages in the UI rather than just logging them, but I don't have the expertise or time to do that. I also couldn't figure out how to instantiate  a `AlarmoCoordinator` object for testing, so there is no unit test coverage.

nielsfaber:
I released your validation PR in the v1.10.10 update, thanks again for your work 👍 
I will keep this issue open as a reminder to myself to add error message to the frontend as well.

nielsfaber:
I added the frontend part for this feature now: https://github.com/nielsfaber/alarmo/commit/fd43037d9585764ab9a4f6906b863f75fbcb1aa7.
The error messages will be shown to the user inside a popup.
I will get back to you when this feature is released.

PR #1211, opened 2025-07-05
Validate user name and code before creating/updating users

therealmrfox:
Proposed fix for [1183](https://github.com/nielsfaber/alarmo/issues/1183).

Note that attempts to create users with conflicting codes or names will silently fail as this pull request doesn't include any UI changes. I've added some logging to give some indication of what's happening.


nielsfaber:
Thank you for the PR, this is a good addition.
I will try to add some feedback to the user in the UI myself.

PR #1257, opened 2025-09-07
Fix `alarmo_failed_to_arm` event docs

k0t3n:
The `sensors` object is actually sent, not `open_sensors`

Example:
```yaml
event_type: alarmo_failed_to_arm
data:
  command: arm_away
  context_id: null
  area_id: "1754766314"
  entity_id: alarm_control_panel.alarmo
  reason: open_sensors
  sensors:
    - binary_sensor.door_sensor_1_contact
    - binary_sensor.door_sensor_2_contact
origin: LOCAL
time_fired: "2025-09-07T19:00:55.453625+00:00"
context:
  id: 01K4JRV10XAGDRZNGF0HY4F3RK
  parent_id: null
  user_id: null
```


nielsfaber:
You are correct, thank you for spotting this and fixing this for everyone 👍

PR #1267, opened 2025-09-26
Create pt.json

alfgomes:
Translate to Portuguese (Brazil)

### DIFF
--- a/custom_components/alarmo/translations/pt.json
+++ b/custom_components/alarmo/translations/pt.json
@@ -1,0 +1,85 @@
+{
+  "services": {
+    "arm": {
+      "name": "Armar",
+      "description": "Armar entidade Alarmo com configura��es personalizadas.",
+      "fields": {
+        "entity_id": {
+          "name": "ID",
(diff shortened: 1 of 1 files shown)


alfgomes:
I couldn't create a Portuguese version of Portugal because there's no way to add both a pt-BR.json and a pt-PT.json, as well as we can create an en-US.json, en-CA.json, en-GB.json, and others!

Or if it's better for you, you can change this file to br.json and I'll create a new one as pt.json!

nielsfaber:
Thank you for your contribution 👍 

> I couldn't create a Portuguese version of Portugal because there's no way to add both a pt-BR.json and a pt-PT.json

Why not? I believe this is the correct approach.
Simple way to test if it works is by setting your account language to brazil / portugal and check that the translations are applied.

Note that this translation is only for the backend part, mostly the custom actions.
For translating the Alarmo configuration panel, a separate translation would need to be added here:
https://github.com/nielsfaber/alarmo/tree/main/custom_components/alarmo/frontend/localize/languages

Issue #686, opened 2023-03-21
Add support to send alarm events to Alarm Company over standarized alarm protocols.


ruant:
### Proposal
Are there plans for adding support to send alarm events to "third party" alarm centrals?
Like adding support for SIA DC-09, or Contact ID?

I would LOVE to get rid of my duplicate sensor setup since the alarm system needs access to it's own set of sensors, and my lighting etc. can't easily get fast enough access to these over the slow cloud. (Especially for motion this doesn't work).

So being able to skip the entire dedicated alarm system and have Alarmo set up in HA and configure it so it sends events to the Alarm company/central over the already standardized protocols that exists for the alarm sector would be awesome.

I know Alarmo aims for a more run your-own alarm, but the added safety of a 3rd party acting on events when I'm not within cell coverage is important to me (and I can't be the only one?)

### Additional info
Perhaps there is a need for a way to include/exclude what sensors should be sent to the external system too? If not all sensors should be acted upon. Like I have some water sensors that isn't really important that the alarm company gets notified about, but it's nice to get notified by Alarmo/HA when they're triggered so I can empty the container it monitors.

dpbarinov:
Hi. I need ability to send alarm events to "third party" alarm centrals too.
Please add this feature!!!

dpbarinov:
Hi. Are you planning to add this feature?

nielsfaber:
@dpbarinov 
> Hi. Are you planning to add this feature?

No. I have no idea about such protocols and I don't think this platform is reliable enough to automatically send reports to an security company.
I would assume such companies require a certified alarm system, rather than some DIY solution with Home Assistant running on a Raspberry Pi.

ruant:
I've been doing a bit more research around it.

And I see that HA has a SIA addon that we maybe can pick some code from.
This HA integration is listen only, to get sensor data into HA. But it contains a lot of the "codes" and what each code means etc.
So in my mind it would be a great starting point since it most likely contains a lot of logic that could be used.
https://www.home-assistant.io/integrations/sia/
https://github.com/home-assistant/core/tree/dev/homeassistant/components/sia
Home Assistant Core is Apache License 2.0, so shouldn't be an issue to snip out some code from here. But not sure what license alarmo is under, can't find any info about it in this repo? @nielsfaber 


SIA is as I understand it the "newest" protocol, so I guess it would be fine to just implement this, and drop ContactID (older protocol).

ruant:
[...]

PR #1270, opened 2025-09-28
feat: add support for sending SIA codes to remote

ruant:
The goal of this PR is to add support to Alarmo to send events/SIA codes to a defined remote "server".


ruant:
Help with development is highly appreciated.

nielsfaber:
Unfortunately, it seems we are not aligned on this feature.
I don't think this implementation should be added to Alarmo.
As discussed in #686, I think it would be better to use the already existing [SIA integration](https://www.home-assistant.io/integrations/sia/) of HA, and focus on a means to connect Alarmo to that instead.
I am closing this PR, you may continue development in a local fork if you want to proceed with this implementation.

PR #1277, opened 2025-10-03
Fix Tabs on 2025.10 (and fix build errors!)

mrtncode:
Hey @nielsfaber ,
In this PR, I fixed the build errors so that the new version could be built with the correct ha-tabs (not sl-tabs). The alarm-panel.js in the dist folder is now functional again (I tested it) and the tabs are working again!

I look forward to hearing back from you soon and to a merge :)
That was a little over three hours of work, and I would be happy if it hadn't been for nothing 😅

P.S. Is there a reason why npm run lint wasn't executed? I was curious when I ran it and had over 30 changed files. But I didn't want to burden this PR with that 😄

### DIFF
--- a/custom_components/alarmo/frontend/rollup.config.js
+++ b/custom_components/alarmo/frontend/rollup.config.js
@@ -24,7 +24,7 @@ export default [
     output: {
       file: 'dist/alarm-panel.js',
       format: 'iife',
-      sourcemap: false,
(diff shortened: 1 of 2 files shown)


jogoodma:
Confirming that applying this PR to my local install of v1.10.11 Alarmo and HA 2025.10.0 has resolved the issue.

Issue #1279, opened 2025-10-04
Entity Unavailable


Switchbot-Nova:
### Checklist

- [x] I am using the latest version of Alarmo (latest version can be found [here](https://github.com/nielsfaber/alarmo/releases/latest))
- [x] I checked for similar existing requests (both open and closed) before posting.

### Alarmo Version

v1.10.12

### HA Version

2025.10.1

### Bug description

Ever since I updated to v1.10.12 the Alarmo entity disappeared from home assistant, going back to v1.10.11 doesnt work either the entity stays unavailable even tho it worked in the past

### Steps to reproduce

Update to v1.10.12

### Relevant log output

```shell
No errors in the logs, havent turned on debug logs yet
```

Siebe-Uy:
More logs:

```

Logger: homeassistant.components.alarm_control_panel
Source: helpers/entity_platform.py:843
integration: Alarm control panel (documentation, issues)
First occurred: 1:00:18 AM (1 occurrence)
Last logged: 1:00:18 AM

Platform alarmo does not generate unique IDs. ID alarm_control_panel.alarmo is already used by alarm_control_panel.alarmo - ignoring alarm_control_panel.alarmo

Logger: homeassistant.util.logging
Source: util/logging.py:167
First occurred: 1:00:19 AM (1 occurrence)
Last logged: 1:00:19 AM

[...]

PR #1283, opened 2025-10-06
Fix listener cleanup, guard against duplicate registration

cayossarian:
Address #1279 by adding guard against duplicate registration and fixing listener cleanup for reloads.

### DIFF
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -39,6 +39,9 @@ from . import const
 
 _LOGGER = logging.getLogger(__name__)
 
+# Store per-config-entry unsubscribe callbacks for platform-level dispatcher listeners
+PLATFORM_UNSUBS = "platform_unsubs"
+
 
 async def async_setup(hass, config):
     """Track states and offer events for alarm_control_panel."""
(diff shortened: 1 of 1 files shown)


nielsfaber:
@cayossarian Could you clarify the reason for closing the PR? Following the discussion in #1279, I was under the impression the changes made are a good addition.

cayossarian:
Yes, thanks for reminding me.  i had some issues with line endings.  But in the meantime I added #1306 which hopefully fixed the unrelated failing test (evidenced by passing CI after prior failure).  

#1306 is the clean update PR for #1279 which also includes the fixed test #1306 as the CI was failing otherwise.  I could close #1306 (or you can if you just want the listener update plus test fix together in this PR).

By the way there was a related discussion in #1279 about unique ID for which you still have the option to programmatically migrate unique ID to something unrelated to Entity ID (vs Unique ID == Entity ID).  This can be done in place, no impact to users, entity ID remains unchanged.  

Practical implication - change the entity ID in the UI and what happens is you get a new copy of the entity with the old one now inaccessible.  Normally, changing entity ID would not create a new entity other than require the user to update their dashboard.  I only bring this up for the sake of awareness.  Something to ponder, if you don't see people having a problem in this area, all good.

PR #1297, opened 2025-10-17
Validation fixes

EVWorth:
lock python to 3.13.2 for homeassistant dependencies to install correctly
uninstall setuptools as it is not needed in the project
bump homeassistant to ~=2025.10.0 to match hacs.json

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,3 +15,6 @@ dev = [
     "pytest-asyncio>=0.26.0",
     "pytest-homeassistant-custom-component>=0.13.243",
 ]
+
+[tool.uv]
+required-version = ">=0.9.0"


EVWorth:
@nielsfaber I wasn't able to reproduce the pillow install issue, but I bumped versions of everything and the uv sync runs correctly now.. however, I am seeing a weird error where a couple of random tests will randomly fail. 
Such as this one in the current run:

FAILED tests/test_ready_to_arm_events.py::test_home_assistant_event_bus_ready_to_arm_event - AssertionError: armed_away should be False when door sensor is open
assert True is False

the strange thing is that when I went to step through the test locally, it worked perfectly fine, and when I re-ran the validate workflow in a test pr against my branch, without having made any other changes, the tests fully passed.

EVWorth:
case in point... I added another small change tgo trigger the validation workflow and this time it worked... 

first workflow failed: https://github.com/nielsfaber/alarmo/actions/runs/18578478705/job/52968515707?pr=1297
second workflow (no relevant changes) succeeded: https://github.com/nielsfaber/alarmo/actions/runs/18578951514/job/52969924975

EVWorth:
Regardless. I think these updates to the pyproject.toml should fix your issues with dependencies.

nielsfaber:
Thanks a lot for helping out with this issue, let's try out if this fixes it.

nielsfaber:
I can confirm the changes are effective 👍 
As you predicted, the issue 
> FAILED tests/test_ready_to_arm_events.py::test_home_assistant_event_bus_ready_to_arm_event - AssertionError: armed_away should be False when door sensor is open

is now preventing the tests from passing, but this seems related to the test itself.
Thank you very much!

EVWorth:
It's weird. If I run tests locally that one fails. If I go into the tests section and run the tests in that specific file those succeed. If I then run all tests again, they all pass

PR #1299, opened 2025-10-17
fix yield to HA in test after advance time

cayossarian:
We needed to add a yield back to HA as the event would work differently on slower/faster systems.

### DIFF
--- a/tests/test_ready_to_arm_events.py
+++ b/tests/test_ready_to_arm_events.py
@@ -1035,6 +1035,8 @@ async def test_home_assistant_event_bus_ready_to_arm_event(
         hass.states.async_set("binary_sensor.test_door", "on")
         await hass.async_block_till_done()
         await advance_time(hass, PROCESSING_TIME)
+        # Ensure all event bus events have been processed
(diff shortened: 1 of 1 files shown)


nielsfaber:
Thanks (as always) for helping out improving the tests!

nielsfaber:
I noticed the fix does not address the issue that is seen in the github automated action, still getting the same error:

> =========================== short test summary info ============================
FAILED tests/test_ready_to_arm_events.py::test_home_assistant_event_bus_ready_to_arm_event - AssertionError: armed_away should be False when door sensor is open
assert True is False
========================= 1 failed, 74 passed in 3.14s =========================
Error: Process completed with exit code 1.

cayossarian:
I'll open a separate issue.  The test does not fail on my system so it's not entirely clear what the root cause is.   May take some trial and error and changes to how we set up the mocks.  @EVWorth do you experience failure with this test?  If so we can work together to try and chase it down.

EVWorth:
Yes but it's really inconsistent

cayossarian:
@EVWorth please try #1306

PR #1305, opened 2025-10-30
Fix listener cleanup, guard against duplicate registration

cayossarian:
Address #1279 by adding guard against duplicate registration and fixing listener cleanup.


nielsfaber:
Thanks for these improvements! I hope these changes will prevent the occassionally reported renaming of the alarm entities, I guess time will tell.

PR #1333, opened 2026-01-06
Create it.json

acasa-source:
Italian (it) localization for the Alarmo integration: translates service names, field labels, descriptions, and arm-mode options in Home Assistant.


nielsfaber:
Thank you for your contribution!

PR #1340, opened 2026-01-10
Fix dropdowns on HA 2026: bundle Vaadin combo-box, bind .items, singl…

Peter0x48:
…e-file build, cache-bust panel


Peter0x48:
Sorry, wrong push

PR #1345, opened 2026-01-12
Add trigger delay setting for alarm sensors

### LINKED ISSUE #1289: Add trigger delay setting for alarm sensors
### Checklist
- [x] My request is generic, other users may benefit from it too.

### Problem
Currently, sensors trigger the alarm (or entry delay) immediately when they switch from "off" to "on". There's no way to require a sensor to remain "on" for a minimum duration before triggering.

### Proposed Solution
Add an advanced/optional delay setting for alarm sensors. The sensor must remain continuously "on" for the configured delay period before triggering the alarm/entry delay.

Default: 0ms (immediate trigger, current behavior)

### Use Case
Prevents false alarms from:
- Momentary sensor activations
- Sensors that briefly trigger but immediately clear
- Scenarios where only sustained activation should trigger the alarm

Example: An AI camera people sensor must detect presence continuously for 0.5 seconds before triggering, ignoring brief glitches. (My actual issue)

jsenecal:
Fixes #1289

This adds a new optional `delay_on` setting for sensors. When configured, the sensor has to stay active for the specified duration before it actually triggers the alarm. If it turns off before the delay expires, nothing happens.

Useful for filtering out momentary glitches from sensors like AI camera detections that might briefly flash on/off.

**Changes:**
- Backend: new `delay_on` field in sensor config, timer management in sensor handler
- Frontend: duration picker in sensor editor (under advanced settings)
- Tests: 9 new tests covering the various scenarios

The delay is applied before the normal entry delay, so they work sequentially (on_delay -> entry_delay -> triggered).


nielsfaber:
Thanks for working on this feature.
Although I am a bit sceptical about adding this function to Alarmo (given the problem caused by the sensor, not the alarm), I understand it's cumbersome to add template binary sensors for time-filtering, so I suppose this is a helpful addition.

I have few comments that I would like to address prior to merging.

**Required change:**
I think the naming `trigger_delay` is a very poor choice. The word 'trigger' is reserved for the 'triggered' state of the alarm itself, which has already a trigger time setting, so this added setting my introduce confusion. I suggest to rename the feature to `delay_on`  (similar to HA's template binary sensor), `time_filter` or anything better.

**Consideration:** 
It would be nice if we can combine the code of this new timer with the already existing arm-timer (which is used for debouncing the sensor signal of closing the front door). Since these are more-or-less the same implementations, the code is redundant/functionally overlapping here. I'm refering to this code: https://github.com/nielsfaber/alarmo/blob/0e3ca1899eee5aaea615490f7787f6ed933ae4b9/custom_components/alarmo/sensors.py#L464-L490

Other than this your code looks very neat, couldn't have done a better job myself.
If I remember correctly, you are the first contributor ever to dare to touch the frontend part of Alarmo, but it seems you found your way through the code.

jsenecal:
Hi @nielsfaber, thanks for the review and kind words!

**Naming:** I do agree that `delay_on` is much better. I don't know why I did not think of that as it does align with HA's template `binary_sensor` naming (`delay_on`/`delay_off`) while avoiding confusion with the _triggered_ state. I'll rename accordingly following this discussion.

**Timer and code reuse:** Fair point, I guess I should have spotted the overlap with the arm-timer. I see two ways to approach this:

**Option A:** Helper methods on `SensorHandler` (Keeps the existing _code_ style, just extracts the common logic)

```python
def _start_entity_timer(
    self,
    timers: dict[str, CALLBACK_TYPE],
    entity: str,
    delay: timedelta,
    callback: Callable[[], None],
) -> None:
    self._stop_entity_timer(timers, entity)
    
    @callback
    def timer_finished(_now):
        timers.pop(entity, None)
        callback()
    
    timers[entity] = async_track_point_in_time(
        self.hass, timer_finished, dt_util.utcnow() + delay
    )

def _stop_entity_timer(self, timers: dict[str, CALLBACK_TYPE], entity: str | None = None) -> None:
    if entity:
        if cancel := timers.pop(entity, None):
            cancel()
    else:
        for cancel in timers.values():
            cancel()
        timers.clear()
```

Usage:
```python
# Arm timer
self._start_entity_timer(self._arm_timers, entity, SENSOR_ARM_TIME, lambda: self._on_arm_complete(entity))
self._stop_entity_timer(self._arm_timers, entity)

# Delay on timer
self._start_entity_timer(self._delay_on_timers, entity, sensor_delay, lambda: self._on_delay_complete(entity))
self._stop_entity_timer(self._delay_on_timers, entity)
```

**Option B:** Separate `EntityTimerManager` class (I think this is cleaner, but it adds a bit more abstraction)

```python
class EntityTimerManager:
    def __init__(self, hass: HomeAssistant):
        self._hass = hass
        self._timers: dict[str, CALLBACK_TYPE] = {}
    
    def start(self, entity: str, delay: timedelta, callback: Callable[[], None]) -> None:
        self.stop(entity)
        
        @callback
        def timer_finished(_now):
            self._timers.pop(entity, None)
            callback()
        
        self._timers[entity] = async_track_point_in_time(
            self._hass, timer_finished, dt_util.utcnow() + delay
        )
    
    def stop(self, entity: str) -> None:
        if cancel := self._timers.pop(entity, None):
            cancel()
    
    def stop_all(self) -> None:
        for cancel in self._timers.values():
            cancel()
        self._timers.clear()
```

Usage:
```python
# In __init__
self._arm_timers = EntityTimerManager(hass)
self._delay_on_timers = EntityTimerManager(hass)

# Arm timer
self._arm_timers.start(entity, SENSOR_ARM_TIME, lambda: self._on_arm_complete(entity))
self._arm_timers.stop(entity)

# Delay on timer
self._delay_on_timers.start(entity, sensor_delay, lambda: self._on_delay_complete(entity))
self._delay_on_timers.stop(entity)
```

This keeps the callback logic in the respective methods while eliminating the duplication but Option A feels more consistent with the existing code.

Does this match what you had in mind? Do you have a preferred approach or were you thinking of something completely different?

Really well-structured project by the way! There is a clear separation between backend, frontend, and the clean state machine made it straightforward to navigate.

nielsfaber:
What I intended to point out, is that a sensor having `ARM_ON_CLOSE` feature, is currently already treated as if having `delay_off` with hard-coded setting `SENSOR_ARM_TIME` equal to 5 seconds. 
I would prefer to have the `start_arm_timer` and `stop_arm_timer` methods removed from the code (together with the `self._arm_timers` struct).
Instead the code would just use your newly added methods for the delay handling, and and call the `_execute_sensor_trigger` when done. Here should then be some logic added that arms the alarm (as is currently done when arm timer is finished).
I don't think there is a need for adding an extra class to achieve this. The only difference is that it filters the on->off state transition (i.e. `delay_off`) rather than the off->on state transition (=`delay_on`), so it would require your code to be adapted to handle opposite state change as well.

Looking at your code, some other consideration came to mind.
I saw that your PR cuts the `async_sensor_state_changed` into two parts (the second part being moved into `_execute_sensor_trigger`).
It is not completely clear to me how this cut was made.
Do you think it is possible to leave the function in one piece, such that all decision logic regarding sensor changes is together?
As alternative, maybe this part:
https://github.com/nielsfaber/alarmo/blob/0e3ca1899eee5aaea615490f7787f6ed933ae4b9/custom_components/alarmo/sensors.py#L221-L223
Could be modified to execute a new function containing the `delay_on` time filtering (if applicable). 
After this filtering is done, it could execute the `async_sensor_state_changed`, so it would just act as in-between delay (delaying off->on state changes only, otherwise just being passthrough).
I may be overlooking something that prevents this from working out, but it feels more structurized to me.
I just wanted to share my thoughts, if you're happy with your implementation it can be merged as-is as well.

matigonzalez98:
thank you! Let’s hope they add this too. I’m having trouble with a window sensor that’s having a 1ms turn on when the sun hits and triggers my alarm ocassionaly. This will fix the issue without requiring to change the sensor!

nielsfaber:
@jsenecal Do have any plan to finish this feature? If adapting the code with my suggestions involves too much extra effort, we can keep it as-is.

jsenecal:
Hi @nielsfaber,

Apologies for the silence - I got sidetracked by the complexity of your proposal and life got the best of me these past few weeks.

Thanks for the clarification. I understand now - you're suggesting generalizing the delay handling to support both `delay_on` (my feature) and `delay_off` (ARM_ON_CLOSE), then replacing the existing arm timer with this unified system.

While I see the appeal of a unified approach, I'd prefer to keep this PR focused on the original scope. A few reasons:

1. **ARM_ON_CLOSE works well as-is** - the hardcoded 5-second delay has been stable and users aren't asking for it to be configurable
2. **Different callbacks** - `delay_on` triggers the alarm via `_execute_sensor_trigger()`, while ARM_ON_CLOSE completes arming via `async_arm()`. Unifying these requires additional abstraction to handle the different outcomes
3. **Risk of regression** - ARM_ON_CLOSE is a well-tested feature; refactoring it alongside a new feature increases the chance of introducing bugs
4. **Scope creep** - this PR addresses a specific problem (#1289). Expanding to refactor ARM_ON_CLOSE changes the scope significantly

That said, I do agree the timer boilerplate is duplicated. As a compromise, I'll implement **Option A** from my earlier comment - extracting helper methods that both timers can use:

```python
def _start_entity_timer(
    self,
    timers: dict[str, CALLBACK_TYPE],
    entity: str,
    delay: timedelta,
    callback: Callable[[], None],
) -> None:
    self._stop_entity_timer(timers, entity)
    
    @callback
    def timer_finished(_now):
        timers.pop(entity, None)
        callback()
    
    timers[entity] = async_track_point_in_time(
        self.hass, timer_finished, dt_util.utcnow() + delay
    )

def _stop_entity_timer(
    self, timers: dict[str, CALLBACK_TYPE], entity: str | None = None
) -> None:
    if entity:
        if cancel := timers.pop(entity, None):
            cancel()
    else:
        for cancel in timers.values():
            cancel()
        timers.clear()
```

This eliminates the duplication while keeping the two features separate. If there's appetite to unify ARM_ON_CLOSE into a configurable `delay_off` feature later, that could be a follow-up PR.

I'll also rename `trigger_delay` to `delay_on` as requested.

Let me know if this approach works for you.

jsenecal:
> I saw that your PR cuts the async_sensor_state_changed into two parts (the second part being moved into _execute_sensor_trigger).
> It is not completely clear to me how this cut was made.
> Do you think it is possible to leave the function in one piece, such that all decision logic regarding sensor changes is together?

The reason `_execute_sensor_trigger` was extracted is that it needs to be called from two places:

1. Directly from `async_sensor_state_changed` when there's no delay configured
2. From the timer callback when the delay expires

If the logic stayed in one function, the timer callback would either need to duplicate the trigger logic or re-run the entire state evaluation from the beginning.

The split creates a clean separation:
- `async_sensor_state_changed` = "should we trigger?" (state validation, filtering, delay management)
- `_execute_sensor_trigger` = "how do we trigger?" (group processing, entry delay calculation, actual trigger call)

I don't see a clean way to keep it in one piece without either duplicating code or adding complexity to pass state through the timer. Open to suggestions if you have a different approach in mind.

nielsfaber:
Thanks again for the work and thought you put into this.
Although I still think that arm-on-close logic could re-use your added code, I understand it may disturb some of the flow, so perhaps it's good to leave it. 
As promised, I will merge your PR as is. After renaming the feature, code looks great to me.
Refinements can always be made later on.

Huh-David:
Nice work here! Any timeline when the next release with this new content will be published?

fencepost-error:
I was wondering why I wasn't seeing this in the settings and noticed it's only enabled for motion and door/window sensors but not smoke/fire sensors which are the ones I'm having trouble with from false triggering.  Could it be enabled for these sensors alongside motion/door/window sensors?  It looks like [this change](https://github.com/nielsfaber/alarmo/pull/1345/changes/b54cfd0d875c694329789e48387d39cdc2e5401e#diff-ca38b9a462ddbbf7db731f159938c69fc97b2964a89185babcb5a866d2191614) would need `ESensorTypes.Environmental` added to the list, but I'm not sure if there are matching changes elsewhere that would also be required, or whether there's a practical reason why they're currently excluded.

Issue #1324, opened 2025-12-01
Alarmo unavailable


piechade:
### Checklist

- [x] I am using the latest version of Alarmo (latest version can be found [here](https://github.com/nielsfaber/alarmo/releases/latest))
- [x] I checked for similar existing requests (both open and closed) before posting.

### Alarmo Version

1.10.13

### HA Version

2025.11.3

### Bug description

The Alarm is unavailable all the time. 

### Steps to reproduce

Try to start the alarm.

### Relevant log output

```shell
2025-12-01 11:25:18.505 DEBUG (MainThread) [custom_components.alarmo.alarm_control_panel] alarm_control_panel.alarmo is removed from hass
2025-12-01 11:25:18.510 DEBUG (MainThread) [custom_components.alarmo.panel] Removing panel
2025-12-01 11:25:18.513 DEBUG (MainThread) [custom_components.alarmo.mqtt] MQTT config was (re)loaded
2025-12-01 11:25:18.515 DEBUG (MainThread) [custom_components.alarmo.alarm_control_panel] alarm_control_panel.alarmo is added to hass
2025-12-01 11:25:18.515 DEBUG (MainThread) [custom_components.alarmo.alarm_control_panel] Initial state for alarm_control_panel.alarmo is unavailable
2025-12-01 11:25:18.515 DEBUG (MainThread) [custom_components.alarmo.alarm_control_panel] entity alarm_control_panel.alarmo was updated from None to unavailable
2025-12-01 11:25:18.516 DEBUG (MainThread) [custom_components.alarmo.alarm_control_panel] Area 1764584337 already registered as alarm_control_panel.alarmo; skipping duplicate add
2025-12-01 11:25:18.518 DEBUG (MainThread) [custom_components.alarmo.alarm_control_panel] ready_to_arm_modes for Alarmo updated to away, home

```

leedrake:
Same issue.  Alarmo running on HASSIO with current version and latest patches for all OS shows as "unavailable".  Dashboard missing.

agherzan:
I have the same (or very similar issue). One of my alarmo areas went "unavailable" today after arming it.

None of the below helped:
- reload
- restart HA
- reboot system
- recreate alarmo area

Renaming the area seems to fix the issue but it is complicated for me as in rely on a specific entity ID.

Debug logs don't seem to provide any relevant information.

Here's some related debug logs

```
[...]

PR #1379, opened 2026-03-05
Fix alarm stuck unavailable after restart

jasonmadigan:
If the last saved state for an alarmo panel is is `unavailable` or `unknown` (e.g. after a Z-Wave stick swap or integration hiccup in my case), the restore logic restores it and the alarm gets stuck permanently. Restarts don't help since the bad state just gets re-saved and re-restored each boot.
                                                                                                                                                    
Small fix: treat unavailable/unknown the same as no previous state, so it falls back to disarmed. This resolves the issue for me.

Found https://github.com/nielsfaber/alarmo/issues/1324 which seems like the same issue.

This may be related too:

https://github.com/nielsfaber/alarmo/issues/1279

### DIFF
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -15,6 +15,8 @@ from homeassistant.util import slugify
 from homeassistant.const import (
     ATTR_NAME,
     ATTR_CODE_FORMAT,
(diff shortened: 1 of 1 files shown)


nielsfaber:
Thanks for helping out, looks like a good improvement 👍

Issue #1386, opened 2026-03-23
Re-include sensors excluded during arming once closed


JanickGers85:
### Checklist
- [x] I checked for similar existing requests (both open and closed) before posting.
- [x] My request is generic, other users may benefit from it too.

### Proposal
Add a feature to automatically re-include sensors that were excluded (bypassed) during the arming phase because they were open, once they transition to the closed state. This would allow arming the alarm while temporarily ignoring open sensors (e.g., with "Bypass automatically" option), but re-enable them in the security system once closed, without needing to disarm and re-arm manually.

### Additional info
Examples:
- A motion sensor active during arming (e.g., home mode) gets excluded, but once it becomes inactive after arming, it gets re-included to monitor future activations.
- A window open during "home" mode activation gets excluded, but once closed, it becomes active again in security monitoring.

Currently, with "Bypass automatically" or "Arm after closing", sensors remain excluded indefinitely.

nielsfaber:
I dont want to offer functions for un-bypassing sensors.
See #1388 for more explanation on the matter.
'Arm after closing' function does not (should not) lead to bypassed sensors.

PR #1388, opened 2026-03-29
Add opt-in auto-reintegration of bypassed sensors on close

nico97118:
# Add opt-in auto-reintegration of bypassed sensors on close

## Summary
This pull request implements an opt-in feature that automatically reintegrates sensors which were bypassed during arming when they return to a safe/closed state. The feature is disabled by default to preserve existing behavior.

**Important branch notes**
- This branch includes two preparatory commits that fix frontend dependencies and linting, required to build the frontend:
  - [480b6ee23b34e9574f5f817fe29b187458f8b293](https://github.com/nielsfaber/alarmo/pull/1388/commits/480b6ee23b34e9574f5f817fe29b187458f8b293) — fix(frontend): complete ESLint 9 and Rollup 4 migration
  - [b81ae75267d1c653a013f3e6b62c6d76e0e2c3ea](https://github.com/nielsfaber/alarmo/pull/1388/commits/b81ae75267d1c653a013f3e6b62c6d76e0e2c3ea) — ESlint after frontend dependancies fixes
- The branch also contains a local custom build artifact commit that might not be merged as-is:
  - [6b49c6ad326a728c61c01693878cbc925cb82391](https://github.com/nielsfaber/alarmo/pull/1388/commits/6b49c6ad326a728c61c01693878cbc925cb82391) — Custom build of alarm-panel.js, should not be merged

*Sadly these 3 commits adds a lot of noise in this PR. I am willing to split the maintenance part ( frontend dependancies and Linting) from the feature if it's easier.*

## What changed

- Backend
  - Add new constant and config attribute: `ATTR_AUTO_REINTEGRATE_BYPASSED_SENSORS` and `Config.auto_reintegrate_bypassed_sensors` (default `False`). See `custom_components/alarmo/const.py`, `custom_components/alarmo/store.py`.
  - Include bypassed sensors in `active_sensors_for_alarm_state` and handle bypass logic in `async_sensor_state_changed`.
  - Implement reintegration logic in `async_sensor_state_changed`: when a bypassed sensor transitions to closed and the flag is enabled, remove it from the area’s `bypassed_sensors` and rebuild listeners so the sensor is monitored again. See `custom_components/alarmo/sensors.py`.
  - Async scheduling when rebuilding listeners so listener registration runs on the Home Assistant event loop (prevents race/context issues).
  - Persist the new config flag in the existing storage load/save logic.

- Frontend
  - Add UI toggle to the General settings panel bound to `auto_reintegrate_bypassed_sensors`. See `frontend/src/views/general/view-general.ts`.
  - Add localization strings for the new UI label and description across languages (`frontend/localize/languages/*.json`).
  - Update websockets validation schema to accept the new boolean field: `custom_components/alarmo/websockets.py`.

- Tests
  - Add/update tests and test factories to cover the opt-in behavior and ensure sensors remain bypassed when the option is disabled. See `tests/test_bypass_mode.py`, `tests/factories.py`,tests/helper.py`.

## Why
Many users (including me) expect bypassed sensors to automatically come back into monitoring when they return to a safe state. Making this behavior opt-in keeps backward compatibility while offering the desired behavior for those who want it. (nielsfaber/alarmo#1386)

## Behavior & migration
- Default: feature disabled (`auto_reintegrate_bypassed_sensors = False`) — no behavior change for existing users.
- When enabled via the UI or API, bypassed sensors will be removed from the bypass list when they close and will immediately be monitored again (and can trigger alarms upon subsequent open).
- No database migrations required beyond storing the new boolean in the existing configuration store.

## How to test (manual)
1. Install the integration from this branch.
2. Enable `auto_reintegrate_bypassed_sensors` in General settings (or via the config API).
3. Arm the alarm (home mode) and bypass a sensor (or allow it to be bypassed during arming).
4. Open the sensor to trigger the alarm; disarm.
5. Re-arm, then close the sensor — check logs for: `removed from bypassed_sensors for area <area_id>`.
6. Re-open the sensor — it should trigger the alarm again.
7. Repeat with the option disabled to confirm sensors remain bypassed on close.

## Files changed
- `custom_components/alarmo/const.py`
- `custom_components/alarmo/store.py`
- `custom_components/alarmo/sensors.py`
- `custom_components/alarmo/websockets.py`
- `frontend/src/views/general/view-general.ts`
- `frontend/localize/languages/*.json`
- `tests/*` (added/updated tests)

## Notes for maintainers
- The frontend dependency/lint fixes are intentionally separated into their own commits to make review easier. They are required to build the new frontend code with modern Rollup and ESLint.
- The local `alarm-panel.js` build artifact commit (`4b7861f`) is included for convenience but should be dropped or re-built in CI if maintainers prefer clean build artifacts from CI.

## Suggested reviewers
- Backend: reviewers familiar with `sensors.py` and `store.py`.
- Frontend: reviewers maintaining `alarm-panel` UI and localization.

## Attachments
- [auto_reintegrate_bypassed_sensor.md](https://github.com/user-attachments/files/26335432/auto_reintegrate_bypassed_sensor.md) — anonymized logs and reproduction steps.


nielsfaber:
I'm sorry, but I don't feel anything for adding the proposed functionality.

The bypassing function (amongst other functions) has been designed by looking at multiple commercial alarm systems (e.g. Honeywell, ADT, DSC).
It is intended to offer a means to arm the alarm in case of a faulty sensor (e.g. empty battery).
I don't want to stimulate or facilitate the regular use (misuse) of sensor bypassing, by offering means to automatically un-bypassing a sensor.
You will not find any such functionality on any of the said alarm systems. 
It is standard practice that a sensor will remain bypassed until the next arming cycle.

In addition, Alarmo does not have even a manual un-bypassing feature (e.g. via alarmo-card or a HA action). This would be a more sensible starting point as it would serve more use-cases and require deliberate action from the user.
However, also for this function I don't see a need at this point.

Please mind the [development guide](https://github.com/nielsfaber/alarmo/blob/main/DEVELOPMENT.md):
> Contributions are much appreciated, but before starting on implementation of a new feature you are advised to create an issue for discussing your plans.

I don't like to receive this PR out of the blue, without any discussion about the subject. 
It looks like you have spend quite a bit of time on this, which I wish to have prevented.
I apologise for turning down your work.

nico97118:
Hi @nielsfaber.
You are right, I should have discussed my plans before sending a PR your way.

As you noticed, I spent some time on this because I wanted to bring the discussion up with a solid base.
I'm not used to contributing to projects so I went too fast in the development and missed the very first step.

I was aware this feature was a bit out of conventional alarm design and that's why I designed it as an advanced opt-in option. But I do understand your point.

For what it's worth, it was motivated by the "idea" to have a kind of "wait and arm" mode : some motions sensor can take a few minutes to clear their state and I don't like to try and arm the alarm several times. This is especially true if I try to "arm after close" on doors, which leads to "several motion sensor stills on" 

Therefore I was trying to achieve a "force the alarm and re-include sensor on safe state", or "wait for safe state before arming". 

> I don't like to receive this PR out of the blue, without any discussion about the subject.

I assume this PR was clumsy since the solution doesn't truly explain the need and I'm sorry if the PR was inappropriate. 
Be assured the intended purpose was to participate on this project which is a superb piece of work I would be happy to contribute to.

Finally would you be interested in the two commits that fix the **dependencies** issues on the frontend I experienced?

nielsfaber:
Thank you for taking the bad news so well.

If you're willing to help with the development, you are most welcome to. 
The best starting point would be to get involved in any of the open feature requests or bug reports, or open new ones to raise attention for any missing or malfunctioning functionality. 
Most people who open an issue don't have the skill to address it themselves, so here you could make a difference.
At the moment I have very limited free time available for the project, my focus is on keeping it up to date with HA and adressing critical bugs.
As said, I prefer if you could give some sort of a heads-up of something you want to work on, so we can align/agree on the direction upfront and your PR doesnt risk being rejected.
I have to admit that I am rather conservative when it comes to adding extra options to the UI, as my aim was/is to keep it simple and understandable for everyone.

> some motions sensor can take a few minutes to clear their state and I don't like to try and arm the alarm several times. This is especially true if I try to "arm after close" on doors, which leads to "several motion sensor stills on"

There is the sensor configuration option 'Allow open initially', which causes the initial open state to be ignored while arming. Subsequent activation would still trigger the alarm. This setting specifically targeted for that purpose. 
Why wouldn't you use this instead?

> Finally would you be interested in the two commits that fix the dependencies issues on the frontend I experienced?

I'm not aware of any issues related to dependencies (meaning package.json?), so if you encountered any and fixed them that would be helpful. My development setup has been running for years so things may be outdated.
Note that from what I could see, your PR introduced a new eslint configuration with different markup rules, which changes the formatting of almost all existing lines of source code. Such PR is rather overwhelming, too many changes to review. I'd prefer to keep the current code as-is, unless you can argue why the change is necessary.

nico97118:
Hi @nielsfaber 

> There is the sensor configuration option 'Allow open initially', which causes the initial open state to be ignored while arming. Subsequent activation would still trigger the alarm. This setting is specifically targeted for that purpose. Why wouldn't you use this instead?

I definitely need to do some tests with this. I missed this option and it looks like the ticket.

> I'm not aware of any issues related to dependencies (meaning package.json?), so if you encountered any and fixed them that would be helpful. My development setup has been running for years so things may be outdated.

Yep, I struggled a bit to get the frontend working on my side due to an incomplete migration from ESLint 8 → 9.
I gave more details in the issue: https://github.com/nielsfaber/alarmo/issues/1389

> Note that from what I could see, your PR introduced a new eslint configuration with different markup rules, which changes the formatting of almost all existing lines of source code. Such PR is rather overwhelming, too many changes to review. I'd prefer to keep the current code as-is, unless you can argue why the change is necessary.

You're perfectly right. I must have failed at some point because after doing the same work a second time, I ended up with a configuration that doesn't mess up every single TS file. As said in the newly opened issue, I prepared a [branch with the fix](https://github.com/nico97118/alarmo/tree/work/eslint9_migration) which I will propose as a PR once you confirm I'm on the right track.

PR #1395, opened 2026-04-05
Issue/1394/refactor_auth

### LINKED ISSUE #1394: Home Assistant freezes for ~6s when disarming (event loop blocked by bcrypt)
### Checklist

### Alarmo Version

1.10.16

### HA Version

2026.4.1

### Bug description

When disarming the alarm, Home Assistant becomes completely unresponsive for approximately 6–9 seconds. No entities update, the UI does not react, and automations are delayed. The alarm does eventually disarm correctly, but HA is fully frozen during the process. The freeze scales with the number of users and areas configured.

This is related to #1177 and PR #1182, but the freeze **persists in v1.10.16** despite the ThreadPoolExecutor fix introduced in that PR. The root cause is that `concurrent.futures.as_completed()` is a blocking call running directly on the HA asyncio event loop thread — the bcrypt checks run in parallel across worker threads, but the event loop itself is blocked waiting for them, which violates the asyncio concurrency model.

Additionally, with multiple areas configured, the master entity calls `alarm_disarm` on each area independently, triggering a full bcrypt authentication cycle for every area — meaning the same code is verified N+1 times against all users (once per area + once for master).

**Setup:**
- 6 users/codes configured
- 4 areas + 1 master
- Code required for both arming and disarming

### Steps to reproduce

1. Install Alarmo and configure at least 3–4 users with PIN codes.
2. Configure at least 2 areas (the more areas, the worse the freeze).
3. Enable the `profiler` integration by adding `profiler` and restarting HA.
4. In **Developer Tools → Services**, call `profiler.start` with `seconds: 15`.
5. Immediately arm the alarm, then disarm it using a valid PIN code.
6. Wait for the profiler to finish. A `.cprof` file will be saved to `/config/`.
7. Analyze it with Python:
   ```bash
   python3 -c "
   import pstats, io
   s = io.StringIO()
   p = pstats.Stats('profile_XXXXXXXX.cprof', stream=s)
   p.sort_stats('tottime')
   p.print_stats(20)
   print(s.getvalue())
   "
   ```
8. Observe `bcrypt._bcrypt.checkpw` and `concurrent.futures.as_completed` at the top of the output, consuming several seconds on the main thread.

**Expected:** Disarming is near-instant from the user's perspective; HA remains responsive throughout.

**Actual:** HA freezes for 6–9 seconds. The freeze duration is proportional to `number_of_users × number_of_areas × bcrypt_time_per_check`.

### Profiler evidence

**By cumulative time:**
```
ncalls  tottime  cumtime  filename:lineno(function)
     5    0.000    9.230   alarmo/alarm_control_panel.py(alarm_disarm)
     5    0.000    9.230   alarmo/alarm_control_panel.py(_validate_code)
     5    0.000    9.230   alarmo/__init__.py(async_authenticate_user)
```

**By internal time:**
```
ncalls  tottime  filename
  30/5    4.231   {built-in method bcrypt._bcrypt.checkpw}
   971    0.142   selectors.py(select)   ← event loop idle, waiting
```

Note: 30 bcrypt calls = 5 disarm calls × 6 users. The same code is verified once per area + once for master.

### Root cause

`async_authenticate_user` in `__init__.py` is not actually async — it's a synchronous function that blocks the event loop:

```python
def async_authenticate_user(self, code, user_id=None):  # not async!
    ...
    with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
        futures = [...]
        for future in concurrent.futures.as_completed(futures):  # BLOCKS event loop
            ...
```

### Proposed fix

**Part 1 — Move authentication off the event loop**

```python
def _authenticate_user_sync(self, code, user_id=None):
    """Blocking authentication — must only be called from an executor."""
    def check_user_code(user, code):
        if not user[const.ATTR_ENABLED]:
            return
        elif not user[ATTR_CODE] and not code:
            return user
        elif user[ATTR_CODE]:
            hash = base64.b64decode(user[ATTR_CODE])
            if bcrypt.checkpw(code.encode("utf-8"), hash):
                return user

    if user_id:
        return check_user_code(self.store.async_get_user(user_id), code)

    users = self.store.async_get_users()
    with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
        futures = [executor.submit(check_user_code, user, code) for user in users.values()]
        for future in concurrent.futures.as_completed(futures):
            if future.result():
                executor.shutdown(wait=False, cancel_futures=True)
                return future.result()

async def async_authenticate_user(self, code, user_id=None):
    """Non-blocking authentication — safe to call from the event loop."""
    return await self.hass.async_add_executor_job(
        self._authenticate_user_sync, code, user_id
    )
```

This also requires making `alarm_disarm` async in `alarm_control_panel.py` and wrapping sync callers with `hass.async_create_task()`.

**Part 2 — Short-lived authentication cache (for multi-area setups)**

With multiple areas, the same code is verified once per area. A short-lived cache (TTL ~5s) would reduce N×M bcrypt calls to just M:

```python
import time, hashlib

_auth_cache: dict = {}
AUTH_CACHE_TTL = 5  # seconds

def _authenticate_user_sync(self, code, user_id=None):
    if not user_id and code:
        cache_key = hashlib.sha256(code.encode()).hexdigest()
        now = time.monotonic()
        if cache_key in _auth_cache:
            result, ts = _auth_cache[cache_key]
            if now - ts < AUTH_CACHE_TTL:
                return result
    result = self._do_authenticate(code, user_id)
    if not user_id and code:
        _auth_cache[cache_key] = (result, time.monotonic())
    return result
```

### Impact summary

| Scenario | Before #1182 | v1.10.16 (after #1182) | With proposed fix |
|---|---|---|---|
| Event loop blocked? | Yes | **Yes** | No |
| Auth calls (4 areas) | 5 × sequential | 5 × parallel threads | 1 × in executor |
| Freeze visible to user | ~9s | ~6s | None |
| With cache (4 areas) | — | — | <0.5s latency, no freeze |

### Environment

- Hardware: 2-core VM

---

I already have a working patch for both parts applied locally and tested. Happy to open a PR if this direction looks good to you.

### Relevant log output

```shell

```

nico97118:
# Authentication Refactor
*This PR Fixes #1394. (Also relates to #1182)*

## TLDR

This refactor removes redundant bcrypt work during master disarm by centralizing code resolution and reusing the resolved user across areas, while preserving all existing permission checks and behaviors.

## Problem Statement

Alarmo validates a user code by iterating over all users and running `bcrypt.checkpw` until a match is found. When disarming **via the master entity**, the code is validated **once per area**. With multiple areas and multiple users, this produces a worst-case cost of:

```
O(n_area * n_user * bcrypt)
```

In addition, `async_authenticate_user` is synchronous on `main` and performs `concurrent.futures.as_completed(...)` directly on the event loop thread. 
The threadpool parallelizes the bcrypt work and **helps** reduce total latency, but the event loop still blocks while it waits for completion. So it improves the symptom without addressing the root cause.

## Solution Design

The refactor separates **authentication** from **authorization** and makes authentication **reusable**:

1. **Resolve code → user once** (authentication, uses bcrypt).
    - bcrypt work is offloaded via `hass.async_add_executor_job` to avoid blocking the event loop.
2. **Validate permissions per area** (authorization, no bcrypt):
   - `area_limit` must include the target area(s)
   - `can_disarm` / `can_arm` must be true for the requested action
3. **Propagate the resolved user** from the master entity to all areas on master disarm.

This ensures that master disarm performs at most **one** bcrypt resolution, while still enforcing per-area permissions.

### 1) `custom_components/alarmo/alarm_control_panel.py`

- Added:
  - `_resolve_user_for_state(...)`: resolves `code → user` once (bcrypt) and validates permissions.
  - `_validate_user(...)`: authorization checks without bcrypt.
  - `_target_area_ids(...)`: helps validate the correct set of areas for master vs area entities.

- `async_alarm_disarm`:
  - Accepts an optional `user` parameter.
  - If `user` is provided, skips bcrypt and only checks permissions.

- `AlarmoMasterEntity.async_alarm_disarm`:
  - Resolves the user **once** at master level.
  - Propagates `user=...` to each area so bcrypt is not re-run.

- Arm actions (`arm_away`, `arm_home`, etc.) already flow through `async_handle_arm_request`, which now uses `_resolve_user_for_state`. This is consistent with the new pattern but the primary gain is for **master disarm**.
- A sync wrapper `alarm_disarm(...)` remains to preserve existing call sites; it now schedules the async implementation.

### 2) `custom_components/alarmo/__init__.py`

- User creation / code changes now hash codes off the event loop via `async_add_executor_job`.
- `async_authenticate_user(...)` is now explicitly async and runs the blocking bcrypt work in an executor.
- `_validate_user_code(...)` and `async_update_user_config(...)` are updated to await the async auth path.
- The previous per-call `ThreadPoolExecutor` was removed to reduce overhead.
- Push-action handlers now `await` the async arm/disarm methods to match the new async signatures.

### 3) `custom_components/alarmo/websockets.py`

- `async_update_user_config(...)` is now awaited, matching its new async signature.

### 4) `tests/test_alarm_master.py`

- Added tests to cover the new auth flow on the master entity:
  - `test_alarm_master_disarm_denied_for_area_limited_user`: a user limited to `area_1` cannot disarm the master when multiple areas are armed.
  - `test_alarm_master_disarm_auth_resolved_once`: master disarm resolves the code only once and reuses the user for all areas.
 
### 5) `tests/factories.py`

- Mock storage includes user CRUD helpers used by the coordinator.

## Additional Information

- **`skip_code=True`** still bypasses validation entirely (previous behavior). This is intentional for internal flows like push actions.
- **Permissions are still enforced per area** even when the user is propagated, ensuring `area_limit` is respected.
- The refactor is focused on **master disarm latency**; area disarm still resolves the code locally, as expected.
- Any external callers using `async_update_user_config(...)` must now await it (the websocket handler was updated accordingly).
- The coordinator includes a thin wrapper `_authenticate_user_sync(...)` that currently just forwards to `_do_authenticate(...)`. It exists to make the “executor-only” contract explicit, but can be collapsed into a single method if reviewers prefer.

### Reproduction steps (thank to @jguillen-lab) :

1. Install Alarmo and configure 5 users with PIN codes.
2. Configure at least 3 areas + master alarm
3. Enable the `profiler` integration.
4. Arm the alarm.
5. In Developer Tools → Services, call `profiler.start` with `seconds: 10`.
6. Immediately disarm using a valid PIN code.

#### Before:

```
    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    20/4    0.625    0.031    0.314    0.078 {built-in method bcrypt._bcrypt.checkpw}
```

#### After:

```
     ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2       0.163    0.081    0.163    0.081 {built-in method bcrypt._bcrypt.checkpw}
```

#### Note: 
- this reflects the shift from worst-case `n_user × n_area` bcrypt checks during master disarm to `n_user` checks max (code resolved once, permissions checked per area).
- difference is not huge in tot time but these tests were run on my development environment which is quite fast. It's better to consider the amount of `ncalls`.

### DIFF
--- a/custom_components/alarmo/__init__.py
+++ b/custom_components/alarmo/__init__.py
@@ -3,7 +3,6 @@
 import re
 import base64
 import logging
-import concurrent.futures
 
 import bcrypt
 from homeassistant.core import (
@@ -49,8 +48,6 @@ from .automations import AutomationHandler
 
 _LOGGER = logging.getLogger(__name__)
 
-# Max number of threads to start when checking user codes.
-MAX_WORKERS = 4
 # Number of rounds of hashing when computing user hashes.
 BCRYPT_NUM_ROUNDS = 10
 
@@ -272,12 +269,14 @@ class AlarmoCoordinator(DataUpdateCoordinator):
 
         async_dispatcher_send(self.hass, "alarmo_sensors_updated")
 
-    def _validate_user_code(self, user_id: str, data: dict):
-        user_with_code = self.async_authenticate_user(data[ATTR_CODE])
+    async def _validate_user_code(self, user_id: str, data: dict):
+        user_with_code = await self.async_authenticate_user(data[ATTR_CODE])
         if user_id:
             if const.ATTR_OLD_CODE not in data:
                 return "No code provided"
-            if not self.async_authenticate_user(data[const.ATTR_OLD_CODE], user_id):
+            if not await self.async_authenticate_user(
+                data[const.ATTR_OLD_CODE], user_id
+            ):
                 return "Wrong code provided"
             if user_with_code and user_with_code[const.ATTR_USER_ID] != user_id:
                 return "User with same code already exists"
@@ -296,7 +295,9 @@ class AlarmoCoordinator(DataUpdateCoordinator):
                 return "User with same name already exists"
         return
 
-    def async_update_user_config(self, user_id: str | None = None, data: dict = {}):
+    async def async_update_user_config(
+        self, user_id: str | None = None, data: dict = {}
+    ):
         """Update user configuration."""
         if const.ATTR_REMOVE in data:
             self.store.async_delete_user(user_id)
@@ -308,7 +309,7 @@ class AlarmoCoordinator(DataUpdateCoordinator):
                 _LOGGER.error(err)
                 return err
         if ATTR_CODE in data:
-            err = self._validate_user_code(user_id, data)
+            err = await self._validate_user_code(user_id, data)
             if err:
                 _LOGGER.error(err)
                 return err
@@ -318,12 +319,17 @@ class AlarmoCoordinator(DataUpdateCoordinator):
                 "number" if data[ATTR_CODE].isdigit() else "text"
             )
             data[const.ATTR_CODE_LENGTH] = len(data[ATTR_CODE])
-            hashed = bcrypt.hashpw(
-                data[ATTR_CODE].encode("utf-8"),
-                bcrypt.gensalt(rounds=BCRYPT_NUM_ROUNDS),
+
+            def _hash_code(code):
+                hashed = bcrypt.hashpw(
+                    code.encode("utf-8"),
+                    bcrypt.gensalt(rounds=BCRYPT_NUM_ROUNDS),
+                )
+                return base64.b64encode(hashed).decode()
+
+            data[ATTR_CODE] = await self.hass.async_add_executor_job(
+                _hash_code, data[ATTR_CODE]
             )
-            hashed = base64.b64encode(hashed)
-            data[ATTR_CODE] = hashed.decode()
 
         if not user_id:
             self.store.async_create_user(data)
@@ -334,32 +340,42 @@ class AlarmoCoordinator(DataUpdateCoordinator):
             self.store.async_update_user(user_id, data)
             return
 
-    def async_authenticate_user(self, code: str, user_id: str | None = None):
-        """Authenticate a user by code."""
+    async def async_authenticate_user(self, code, user_id=None):
+        """Run user authentication in the executor.
 
-        def check_user_code(user, code):
-            """Returns the supplied user object if the code matches, None otherwise."""
-            if not user[const.ATTR_ENABLED]:
-                return
-            elif not user[ATTR_CODE] and not code:
-                return user
-            elif user[ATTR_CODE]:
-                hash = base64.b64decode(user[ATTR_CODE])
-                if bcrypt.checkpw(code.encode("utf-8"), hash):
+        This wrapper ensures that the blocking bcrypt-based authentication
+        is executed outside of the event loop, making it safe to call
+        from async context.
+        """
+
+        def _authenticate_user_sync():
+            """Perform user authentication using blocking bcrypt operations."""
+
+            def check_user_code(user, code):
+                """Returns None or the supplied user object if the code matches."""
+                if not user[const.ATTR_ENABLED]:
+                    return None
+                elif not user[ATTR_CODE] and not code:
                     return user
+                elif user[ATTR_CODE]:
+                    try:
+                        hash = base64.b64decode(user[ATTR_CODE])
+                    except Exception:
+                        return None
+                    if bcrypt.checkpw((code or "").encode("utf-8"), hash):
+                        return user
 
-        if user_id:
-            return check_user_code(self.store.async_get_user(user_id), code)
+            if user_id:
+                return check_user_code(self.store.async_get_user(user_id), code)
 
-        users = self.store.async_get_users()
-        with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
-            futures = [
-                executor.submit(check_user_code, user, code) for user in users.values()
-            ]
-            for future in concurrent.futures.as_completed(futures):
-                if future.result():
-                    executor.shutdown(wait=False, cancel_futures=True)
-                    return future.result()
+            users = self.store.async_get_users()
+            for user in users.values():
+                result = check_user_code(user, code)
+                if result:
+                    return result
+            return None
+
+        return await self.hass.async_add_executor_job(_authenticate_user_sync)
 
     def async_update_automation_config(
         self,
@@ -421,21 +437,21 @@ class AlarmoCoordinator(DataUpdateCoordinator):
 
             if action == const.EVENT_ACTION_FORCE_ARM:
                 _LOGGER.info("Received request for force arming")
-                alarm_entity.async_handle_arm_request(
+                await alarm_entity.async_handle_arm_request(
                     arm_mode, skip_code=True, bypass_open_sensors=True
                 )
             elif action == const.EVENT_ACTION_RETRY_ARM:
                 _LOGGER.info("Received request for retry arming")
-                alarm_entity.async_handle_arm_request(arm_mode, skip_code=True)
+                await alarm_entity.async_handle_arm_request(arm_mode, skip_code=True)
             elif action == const.EVENT_ACTION_DISARM:
                 _LOGGER.info("Received request for disarming")
-                alarm_entity.alarm_disarm(None, skip_code=True)
+                await alarm_entity.async_alarm_disarm(None, skip_code=True)
             else:
                 _LOGGER.info(
                     "Received request for arming with mode %s",
                     arm_mode,
                 )
-                alarm_entity.async_handle_arm_request(arm_mode, skip_code=True)
+                await alarm_entity.async_handle_arm_request(arm_mode, skip_code=True)
 
         self._subscriptions.append(
             self.hass.bus.async_listen(const.PUSH_EVENT, async_handle_push_event)
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -320,8 +320,13 @@ class AlarmoBaseEntity(AlarmControlPanelEntity, RestoreEntity):
             "last_triggered": self.last_triggered,
         }
 
-    def _is_code_required(self, to_state):
-        """Check if code validation is required based on config."""
+    async def _validate_code(self, code, to_state):  # noqa PLR0911
+        """Validate code and user permissions for a requested state change.
+
+        Returns a (success, error_event) tuple. When success is True,
+        error_event is None.
+        """
+        # check bypass rules
         if (
             to_state == AlarmControlPanelState.DISARMED
             and not self._config[const.ATTR_CODE_DISARM_REQUIRED]
@@ -329,7 +334,8 @@ class AlarmoBaseEntity(AlarmControlPanelEntity, RestoreEntity):
             _LOGGER.debug(
                 "Code is not required for disarming, bypassing code validation."
             )
-            return False
+            self._changed_by = None
+            return True, None
 
         if (
             to_state != AlarmControlPanelState.DISARMED
@@ -337,7 +343,8 @@ class AlarmoBaseEntity(AlarmControlPanelEntity, RestoreEntity):
             and not self._config[ATTR_CODE_ARM_REQUIRED]
         ):
             _LOGGER.debug("Code is not required for arming, bypassing code validation.")
-            return False
+            self._changed_by = None
+            return True, None
 
         if (
             AlarmControlPanelState.DISARMED not in {to_state, self._state}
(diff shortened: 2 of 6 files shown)


nielsfaber:
Thank you for working on this.

I do agree that it is unnecessary to repeat the authentication for each area when sending an arm/disarm command to the master, it can be centralized instead.

However, I don't fully agree with this refactoring PR. 
Refactoring is usually intended to clean up code, but I see quite some lines added here which have the opposite effect.
I will add some code review to start the discussion, hopefully we can converge to a good solution.

custom_components/alarmo/__init__.py
```diff
@@ -334,32 +340,42 @@ def async_update_user_config(self, user_id: str | None = None, data: dict = {}):
...
-                if future.result():
-                    executor.shutdown(wait=False, cancel_futures=True)
-                    return future.result()
+        for user in users.values():
+            result = check_user_code(user, code)
+            if result:
+                return result
+        return None
+
+    def _authenticate_user_sync(self, code, user_id=None):
```
nielsfaber: change to nested function within async_authenticate_user (assuming this is the only place this is used)?
I don't understand why we need a sync + async version of the same method.

custom_components/alarmo/__init__.py
```diff
@@ -334,32 +340,42 @@ def async_update_user_config(self, user_id: str | None = None, data: dict = {}):
...
-            futures = [
-                executor.submit(check_user_code, user, code) for user in users.values()
-            ]
-            for future in concurrent.futures.as_completed(futures):
-                if future.result():
-                    executor.shutdown(wait=False, cancel_futures=True)
-                    return future.result()
+        for user in users.values():
+            result = check_user_code(user, code)
+            if result:
```
nielsfaber: shouldn't the loop be stopped once a match is found?

custom_components/alarmo/__init__.py
```diff
@@ -334,32 +340,42 @@ def async_update_user_config(self, user_id: str | None = None, data: dict = {}):
             self.store.async_update_user(user_id, data)
             return
 
-    def async_authenticate_user(self, code: str, user_id: str | None = None):
-        """Authenticate a user by code."""
+    def _do_authenticate(self, code, user_id=None):
```
nielsfaber: I think the old naming of this function was better/more meaningful.

custom_components/alarmo/alarm_control_panel.py
```diff
@@ -1403,18 +1427,39 @@ def async_update_state(self, state: str | None = None):  # noqa PLR0912, PLR0915
...
+
+        res = await super().async_alarm_disarm(
+            code=code, skip_code=skip_code, user=user, context_id=context_id
+        )
         if res:
             for item in self.hass.data[const.DOMAIN]["areas"].values():
                 if item.state != AlarmControlPanelState.DISARMED:
-                    item.alarm_disarm(code=code, skip_code=skip_code)
+                    await item.async_alarm_disarm(
+                        code=code, skip_code=skip_code, user=user
```
nielsfaber: shouldn't we set skip_code to True to have the authentication skipped for all areas?
I was under the impression that this was the main purpose of this PR.

nielsfaber (changes requested):

custom_components/alarmo/alarm_control_panel.py
```diff
@@ -320,79 +320,90 @@ def extra_state_attributes(self):
...
             return (False, const.EVENT_INVALID_CODE_PROVIDED)
         else:
-            self._changed_by = res[ATTR_NAME]
-            return (True, res)
+            self._changed_by = user[ATTR_NAME]
+            return (True, user)
 
-    @callback
-    def async_service_disarm_handler(self, code, context_id=None):
+    async def _resolve_user_for_state(self, code, to_state, user=None):
```
nielsfaber: I dont understand why this function is added.
The logic moved into here should be/remain part of code validation.
The name of this function doesn't help understanding its purpose (description is unclear to me).

nielsfaber (changes requested):

custom_components/alarmo/alarm_control_panel.py
```diff
@@ -320,79 +320,90 @@ def extra_state_attributes(self):
...
-            and not self._config[const.ATTR_CODE_MODE_CHANGE_REQUIRED]
-        ):
-            self._changed_by = None
-            return (True, None)
-        elif not code or len(code) < 1:
-            return (False, const.EVENT_NO_CODE_PROVIDED)
-
-        res = self.hass.data[const.DOMAIN]["coordinator"].async_authenticate_user(code)
-        if not res:
+    def _target_area_ids(self):
```
nielsfaber: I think this function is unnecessary, as for an area the output is evident. On top of that, this function is only used/called from a single place in the code, why not use the if/else there?

nielsfaber (changes requested):

custom_components/alarmo/__init__.py
nico97118: That is the case here.
When a match is found, result is not none, then : 
```python
if result:
    return result
```
stop the loop

custom_components/alarmo/__init__.py
nico97118: I did not rename the function, the diff is a bit messy.
The function `async_authenticate_user` still exists at line R374.

However it now really is an async method

custom_components/alarmo/__init__.py
nico97118: I understand that it's intriguing, explained in the PR notes : 
> The coordinator includes a thin wrapper _authenticate_user_sync(...) that currently just forwards to _do_authenticate(...). It exists to make the “executor-only” contract explicit, but can be collapsed into a single method if reviewers prefer.

custom_components/alarmo/alarm_control_panel.py
nico97118: I thought this would make the code easier to understand.
The idea was to add an helper following the Single Responsibility Principle.

But we can definitely go back to previous with the if/else statement in `_validate_user(...)`

custom_components/alarmo/alarm_control_panel.py
nico97118: > The logic moved into here should be/remain part of code validation.

Actually, I really tried to separate things.
- We want to validate code only once (to reduce bcrypt work), that's handled by the current `_do_authenticate(...)`
- `_validate_user(...)` checks that given user has permissions to arm/disarm for the area we are working in. It does not resolve code -> user. It does not handle global area configuration (.e.g `self._config[const.ATTR_CODE_DISARM_REQUIRED]`) 
- Finally `_resolve_user_for_state(...)` handle the auth for the area/master : 
  - if `user` is given, check its permission via `_validate_user` (useful for recursive disarm from master to avoid resolving code again)
  - if no user were given nor `_validate_user(...)` were able to check permissions for the area, check the area configuration (.e.g `self._config[const.ATTR_CODE_DISARM_REQUIRED]`)
  - Otherwise, no user were given AND area needs code for arm/disarm (typically, the arm/disarm command was called on the area directly so user was not resolved by master) so we call `async_authenticate_user(..)` (-> `_do_authenticate(...)`) followed by `_validate_user(...) 

At the end, this method allow to handle authentication `recursively` whether it was called on master or directly to an area.

I hope these explanations make sense

custom_components/alarmo/alarm_control_panel.py
nico97118: From PR  section Additional Information :

> skip_code=True still bypasses validation entirely (previous behavior). 

>  Permissions are still enforced per area even when the user is propagated, ensuring area_limit is respected.

From my understanding, the behavior when using `skip_code` does not only bypass code resolving. It actually bypass every permissions checks and I tried to keep this the less disruptive as I can. So I used the new `user` argument to avoid a potential security breach. (i.e. skip_code is propagated without making sure the targeted area is allowed to user first)

nico97118:
Hi @nielsfaber,
Thank you for reviewing this. To be honest, I’m not familiar with the GitHub review UI. So apologize if I do any mistake while handling changes

custom_components/alarmo/alarm_control_panel.py
nico97118: After digging deeper in the code. I think I got your point.

```python
res = await super().async_alarm_disarm(
            code=code, skip_code=skip_code, user=user, context_id=context_id
```
This technically already check user permissions for  desired state (`disarmed`) and all areas.
So I guess you're right and we could probably blind trust this res and set "skip_code" to `true` inside the loop
```python
 for item in self.hass.data[const.DOMAIN]["areas"].values():
```
Good catch

nico97118:
Concerning the new helpers methods:
- _is_code_required(...)
- _is_user_allowed_for_area(...)
- _resolve_user(...)
-  _authenticate_user_sync(...)

 They could technically be nested, since they’re only used by the wrapper/pipeline. I kept them as class methods on purpose for readability, testability, and to keep the async wrapper focused on scheduling. 
That’s a style choice though - I will nest them if the next commits

nico97118:
I’ve added a few commits without rewriting the Git history to avoid disrupting the current review process.

However, once we align on the changes, I’d like to clean up the branch (e.g., squash/reorganize commits) so we can merge a clearer and more maintainable history.

@nielsfaber Would it be possible to leave me a bit of time to do this before merging, once everything is approved?

nielsfaber:
Thank you for considering my review comments and improving your code.
It looks quite cleaner and more understable to me.
 
I have only one remark left.
I see you added some nested functions `_is_code_required`, `_is_user_allowed_for_area`, `_is_user_allowed_for_action`, `_resolve_user` in the code validation part of the alarm entities.
As far as I could see, these are only called from a single point in the code, so it seems that you could also just insert the code within the nest functions at the places where they are called.
In other words: there seems no benefit in the nested functions, since they are not re-used anywhere.
I would prefer to not have any sub-functions created unless required, since they require jumping across code when reviewing. For readability I prefer flat code (which is the format as it was).
Could you look into this?

nico97118:
Thanks for the clarification. I’ve pushed a commit that flattens _validate_code as requested.

That said, I’d like to note that I’m not fully convinced by a strict ‘no helper if used once’ rule.
In my opinion these helpers should exists as internal class method as they still bring value (even if used once):
- they keep responsibilities separated (SRP) and make the logic easier to scan,
- they can be unit‑tested independently for non‑regression,
- they reduce the risk of duplication when similar checks appear elsewhere,
- and if reuse becomes necessary later, having helpers already avoids refactoring.

A concrete example: `_validate_code()` handles multiple concerns in one flow (bypass rules, user resolution, area permissions, action permissions). 
The method is quite long and branching-heavy, which led to add the ` # noqa PLR0911` to bypass ruff for multiple returns. While it is possible to workaround this by removing some `return` this is an indirect symptom of the complexity I mentioned.

I understand you prefer flat code for review readability, so I’m aligning with your style here, but I’d still advocate for small helpers when they improve clarity, responsibility, testability and maintainability.

custom_components/alarmo/__init__.py
nico97118, line 539: Update : the lint is broke since #1399 
Fix proposed at #1403 

I don't understand why ruff is suddenly rising an error on this line.
I didn't change anything here.
I have no error running ruff check locally.. and obviously this line is 59 char longs and not 93 as mentioned by the validation workflow. No idea what happened here

nico97118:
Rebased on main to be up-to-date with main's change. (push force to [50739b8](https://github.com/nielsfaber/alarmo/commit/50739b8596865ecb7a389d3d3ad1255ff62d46d8))
Also did some cleanup in commits (push force to [41058ea](https://github.com/nielsfaber/alarmo/commit/41058ea0f7cf6cd495d14bd4e3c2f3eae44a15b2))

I did two independent push to make it easier to review. 
First one: just introduce the diff on main since last rebase.
Second one: is empty since it is only commits squash, reorder, renaming.

nielsfaber:
Thank you for spending the extra effort of restructing the code to my preferences.
My apologies for being picky about the formatting.
I agree that the code may benefit from having some helpers and increased structure.
Particularly the `alarm_control_panel.py` file has been growing over the years, to >1500 lines now. 
If you have some ideas how this could be reduced (while keeping the core alarm class in one file), please share your thoughts.
There is already a helpers.py file but ironically it is almost empty, I never found a good way to introduce separation.

PR #1401, opened 2026-04-13
ESLint 9 migration fixes

nico97118:
Since https://github.com/nielsfaber/alarmo/commit/1f25cf8437a6372800a53924c7786c104209b887
I cannot build the frontend of the app.
This PR complete the ESLint9 and Rollup4 migration by transposing the previous behavior via the new config files.

I didn't add the prettier plugin since it implies a lot of modifications in the src/**.ts files
It could be a good idea to add it at some point though

This PR should fixes  #1389

PR #1402, opened 2026-04-13
workflow: add frontend build and linter check to GitHub Actions

nico97118:
While fixing the frontend dependancies issues. I think it would be a good idea to have some automatic workflow that run linter checks, and even check that the frontend is still buildable.

This PR adds such a workflow, however, since the frontend is not buildable right now on main branch, it would be better to merge this PR after #1401

### DIFF
--- a/.github/workflows/frontend-build.yaml
+++ b/.github/workflows/frontend-build.yaml
@@ -1,0 +1,33 @@
+name: Frontend Build Check
+
+on:
+  pull_request:
+    paths:
+      - 'custom_components/alarmo/frontend/**'
+  push:
+    paths:
+      - 'custom_components/alarmo/frontend/**'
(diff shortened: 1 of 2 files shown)


nielsfaber:
The github action seems to fail since it cannot find the eslint config file.
Since PR #1401 is now merged, I wouldn't expect the action to fail.
I manually triggered a re-run the action just now, it doesn't resolve the issue.
Do you have any clue what's going on here?

nico97118:
Hi @nielsfaber
Thanks for considering my PR.
Indeed the workflow was still failing since my branch was based on main before the #1401 

Just rebased it on main now the ESLint migration is fixed ;)

nielsfaber:
Thanks for the quick fix 👍

PR #1399, opened 2026-04-11
Fix startup/reload failures - dispatcher crash, unique ID collisions, stale config-entry callbacks

codyc1515:
This PR addresses three separate root causes seen during startup/reload, each fixed independently and committed separately:

1. Guard dispatcher sends when entities are detached  
2. Make Alarmo entity `unique_id` truly stable and collision-safe  
3. Clean up handler subscriptions on unload and block stale entity registration callbacks

Fixes #1398. I have tested this locally on my HASS installation. The integration, in its current state, could not even create an alarm entity for me (it was very broken).

## Problem
The integration could fail in three ways:

- `AttributeError: 'NoneType' object has no attribute 'loop'` from `dispatcher_send(...)` after entity detach/unload.
- Duplicate entity warnings: Alarmo entities reported as not generating unique IDs, causing entity registration to be ignored.
- `HomeAssistantError: Can't link device to unknown config entry ...` when stale callbacks attempted to add entities after unload/reload timing.

### 1) Detached-entity dispatcher guard
- Added safety checks in `update_ready_to_arm_modes(...)` for both area and master entities.
- If `self.hass` or `self.hass.loop` is unavailable, skip dispatch and log debug instead of crashing.

Commit: `ae3caca`

### 2) Stable unique IDs + entity ID collision handling
- Introduced deterministic unique IDs based on coordinator ID + area/master identity.
- Replaced `unique_id` behavior that depended on `entity_id`.
- Added entity-id allocation logic that avoids collisions with existing states, registry entries, and already-instantiated Alarmo entities.
- Preserves existing registry mapping when a matching `(domain, platform, unique_id)` already exists.

Commit: `5a484f5`

### 3) Proper unload lifecycle + stale callback guards
- Added explicit `async_unload()` cleanup methods for Sensor, Automation, MQTT, and Event handlers.
- Coordinator now calls handler unload methods during integration unload.
- Guarded area/master registration callbacks to no-op when the config entry is no longer present.
- Registered master entity dispatcher listeners with `self.async_on_remove(...)` to ensure listener cleanup.

Commit: `2e48a85`

PR #1403, opened 2026-04-13
fix(lint): fix linter error introduced by 2e48a853e

nico97118:
Small lint error introduced via #1399 (2e48a853e)


nico97118:
This PR is out of date since the code it was fixing has been reverted 1b0dab2

PR #1407, opened 2026-04-15
[feature] Alarmo-table Sort and Search

### LINKED ISSUE #1319: Sort Notifications and Actions in Alarm Panel configuration
Hi,

I would find it helpful if notifications and tasks were sorted by name by default in the panel list, or if a sorting function were added. Sorting would be generally useful with a large number of notifications and tasks, and sorting by name would be ideal, in my opinion.
Many thanks.

The screenshot shows that the tasks marked in red and green are not sorted and therefore not listed together.

Image

nico97118:
# Add sort button and search input on relevant columns in Frontend
*This PR fixes #1319*
@nielsfaber This is a draft since this feature sounds like a good UX improvement in my opinion. I'm totally open to discuss the implementation.

## TLDR
This PR introduces a generic table sorting and search feature for Alarmo tables, enables sorting in the Actions, Codes, and Sensors panel.

## Why
- The panels needed basic sorting (name/enabled) to make large lists easier to scan.
- Search input is also a huge helper for big lists.

### Panel sorting
- Added a generic sorting mechanism in `alarmo-table`..
- Enabled sorting on **Name** and **Enabled** columns in the Actions panel.
- Added raw fields (`raw_name`, `raw_enabled`) to sort on actual values instead of rendered HTML. (for Codes)
- Active sensors table: sortable by entity name and enabled state.
- Add-sensors table: sortable by entity name.
- Minor width adjustment for the `enabled` column to reduce crowding.

### Sorting UI feedback
- The active sort column is highlighted (icon color + opacity). Inactive sort icons are dimmed.
- Implemented in `alarmo-table` by adding `active/inactive` classes to the sort button and icon.

### Panel Search
- Added a generic search method `matchesSearch` based on 
- Adapt `renderFilterRow` to also filter on search input (via `matchesSearch`)
- Dynamically extends the filter-div to display the search `ha-input-field` if at least one column is `searchable`

## UX Details
- Active sort: `--primary-color`
- Inactive sort: reduced opacity
- Sorting is triggered via the header icon; there is no change to the underlying data model or backend behavior.
- Search is simpler, it is basically just an ha-input-field monitored by search mechanism

## Testing
- Manual: verify sort icons show active/inactive state and sorting works for:
  - Actions: name, enabled
  - Codes (Users): name, enabled
  - Sensors (active): name, enabled
  - Sensors (add): name
- Manual: verify search function

## Screenshots

Capture d’écran 2026-04-16 à 00 23 02
Capture d’écran 2026-04-16 à 00 24 36

## Notes
- Sorting relies on raw values for Users because table cells contain HTML templates.


nielsfaber:
Thanks a lot for spending effort working on this!
I'm impressed with the implementation, it works well and its seamless integrated in the existing code.
However, I do have quite some remarks that ask for refinement. I am hoping that we could discuss a bit this a bit more before proceeding.

1. I see that the functionality is added on all tables, except for the areas table on the general page. Is this deliberate?
2. I think the search bar is quite dominant in the UI, it is rather distracting. I expect most users will not use it so it should be calling less attention.
3. For some tables, like on the codes page, I think the search bar is completely unnecessary since I don't expect the list to grow longer than 10 items in a users setup. I would mostly use it for the sensors and actions pages.
4.  I noticed the sorting function is volatile, i.e. a user would have to apply his preferred sorting every time he accesses the page. I see that sorting in HA is non-volatile. I don't know how they managed to do this, but it is more intuitive.
5. The sorting function is only applied to the main column (the name) and the enabled state. Why not on the other columns? In relation to this, I think we should stretch the width of the tables to the full screen width, and introduce more columns. for example we could add for the sensors a 'type' and 'area' column. This seems more in-line with HA and makes the sorting function more useful.
6. The sorting arrow icon buttons are too small. Its not touch-screen friendly. I think around 24x24 pixels used as minimum within HA.
7. I'm missing an X button in the filter text field for erasing an applied filter.

In relation to comment 2, 6, 7 I think we can use HA's own table design as example:
image

In relation to comment 5, I think some more preparation work would be needed, not directly related to this PR. 
We can certainly consider this as 'roadmap feature' and park this for now.
I have been thinking getting rid of the grey sidebars in the UI for a longer time. Mostly on the sensors and actions page the imposed limitation of width just limits the overview. I see that within HA, pages with large tables fill the complete page, and I think we should follow that example. In both the sensors and actions pages, there are currently 2 separate tables below eachother. I think there are ways to merge them to create a large unified table view.

Please share your thoughts about my remarks.
Hope you are OK with keeping the PR open a bit longer.

nico97118:
> I see that the functionality is added on all tables, except for the areas table on the general page. Is this deliberate?

That's an oversight on my side. I assumed users probably wouldn't have enough areas to really need search/sorting there.

> I think the search bar is quite dominant in the UI, it is rather distracting. I expect most users will not use it so it should be calling less attention.

I totally agree with this point. To be honest, frontend/UI is not really my area of expertise. I tried reducing its size a bit, but I didn't manage to do it properly and eventually lost patience 😅

> For some tables, like on the codes page, I think the search bar is completely unnecessary since I don't expect the list to grow longer than 10 items in a user's setup. I would mostly use it for the sensors and actions pages.

On this point, I don't fully agree. I understand most users might only use a single pincode, but I genuinely think some setups can become quite large with RFID tags, temporary codes, guest access, etc.

For example, imagine a family with two parents and three children using RFID tags for convenience and individual pincodes as backup. That's already 10 codes. Add a temporary code for a housekeeper or another family member, and the list starts becoming less manageable.

I personally like using individual auth to track which users interacted with Alarmo. On my side, every permanent user has an RFID tag, a strong password (mainly used through smartphone shortcuts), and a keypad pincode. That's already 3 codes per users

I'll let you decide whether this use case is worth considering.

> I noticed the sorting function is volatile, i.e. a user would have to apply his preferred sorting every time he accesses the page. I see that sorting in HA is non-volatile. I don't know how they managed to do this, but it is more intuitive.

I don't have a strong opinion on this yet. I'll check whether there's a clean way to store sorting preferences in some kind of user cache or local storage.

> The sorting function is only applied to the main column (the name) and the enabled state. Why not on the other columns? In relation to this, I think we should stretch the width of the tables to the full screen width, and introduce more columns. for example we could add for the sensors a 'type' and 'area' column. This seems more in-line with HA and makes the sorting function more useful.

I agree with this point. I mainly wanted to avoid mixing multiple goals within the same PR.

My thinking was that it would be better to first merge a simpler and well-scoped implementation of the feature, then improve the overall table UI later in a separate PR using the newer mechanisms available.

> The sorting arrow icon buttons are too small. It's not touch-screen friendly. I think around 24x24 pixels used as minimum within HA.

I actually struggled with the CSS while trying to reduce their size 😆

But you're right: the default HA button was definitely more touch-friendly.

Again, frontend design is not really my strongest area.

> I'm missing an X button in the filter text field for erasing an applied filter.

Good point. I need to look into how to add that, but I expect it shouldn't be too complicated.

> Mostly on the sensors and actions page the imposed limitation of width just limits the overview. I see that within HA, pages with large tables fill the complete page, and I think we should follow that example.

100% agree. But again, this is the kind of change I'd prefer to address in a separate PR to avoid mixing multiple discussions and objectives here.

I designed this PR to make the added functionality easy to review while still being usable enough to merge as-is. I think it could already improve the experience for some users (e.g. #1319).

That said, it's completely up to you if you'd rather first redesign the table UI more broadly and introduce these features afterward. I'd be happy to help in both cases 😉

> In both the sensors and actions pages, there are currently 2 separate tables below each other. I think there are ways to merge them to create a large unified table view.

Same remarks as above, but here are a few additional thoughts:

- Regarding the sensors page, I think merging both tables could become confusing. However, moving the "Add sensors" section into a dedicated modal/popup (similar to the UX used for adding users or actions) could improve the experience.
- Regarding notifications/actions, I also think merging everything into a single table might become confusing. However, splitting them into two tabs ("Notifications" and "Actions") could be worth considering.

----

Overall, this PR was mainly intended to start the discussion around these features, so I'm perfectly fine if it ends up being parked for a while.

On the other hand, I should mention that I intentionally didn't spend additional time adapting everything perfectly to every field because I first wanted feedback on the overall direction.

The mechanisms are implemented in a scalable way, so adding or removing search/sorting support on specific columns is straightforward. Both features are token-driven through the column declarations.

I enabled them where they made the most sense to me as a proof of concept, but it's quite easy to expand or reduce the searchable/sortable fields.

Same logic for sorting support. Please note that the current sorting implementation handles numeric values and otherwise falls back to string sorting. Additional adjustments might be needed if you want sorting support for more complex data types such as lists (e.g. sensor types).

----

Looking forward to hearing your thoughts.

nico97118:
I significantly reworked this PR. It is now based on #1425, which is itself based on #1424. I genuinely wanted to make this a stacked PR, but I couldn't figure out how to do that cleanly on GitHub.

Sorry for the larger review this creates.
I tried to keep the commit history tidy so the changes can hopefully be reviewed step by step.

Here are a few screenshots from my sandbox setup:
Capture d’écran 2026-05-18 à 00 32 49
Capture d’écran 2026-05-18 à 00 32 23
Capture d’écran 2026-05-18 à 00 32 06

xblitz:
Wow this is a great PR @nico97118 , well made and a lot of effort into taking feedback and post an update, @nielsfaber could this be reviewed ?

Issue #1373, opened 2026-02-18
Error in Home Assistant log


uvjim:
### Checklist

- [x] I am using the latest version of Alarmo (latest version can be found [here](https://github.com/nielsfaber/alarmo/releases/latest))
- [x] I checked for similar existing requests (both open and closed) before posting.

### Alarmo Version

v1.10.15

### HA Version

2026.2.2

### Bug description

Spotted these logs in my Home Assistant log. I'm not sure when they started and also I'm not sure how to reproduce them.

```
Logger: homeassistant.helpers.dispatcher
Source: util/logging.py:167
First occurred: 17 February 2026 at 22:07:15 (27 occurrences)
Last logged: 12:12:28

Exception in async_handle_event when dispatching 'alarmo_event': ('ready_to_arm_modes_changed', None, {'modes': []}) Traceback (most recent call last): File "/config/custom_components/alarmo/event.py", line 74, in async_handle_event supported_modes = dict( filter( ...<2 lines>... ) ) File "/config/custom_components/alarmo/event.py", line 76, in <lambda> lambda el: el[1] & alarm_entity.supported_features, ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ AttributeError: 'NoneType' object has no attribute 'supported_features'
Exception in async_handle_event when dispatching 'alarmo_event': ('ready_to_arm_modes_changed', None, {'modes': [<AlarmControlPanelState.ARMED_AWAY: 'armed_away'>, <AlarmControlPanelState.ARMED_HOME: 'armed_home'>, <AlarmControlPanelState.ARMED_NIGHT: 'armed_night'>, <AlarmControlPanelState.ARMED_VACATION: 'armed_vacation'>]}) Traceback (most recent call last): File "/config/custom_components/alarmo/event.py", line 74, in async_handle_event supported_modes = dict( filter( ...<2 lines>... ) ) File "/config/custom_components/alarmo/event.py", line 76, in <lambda> lambda el: el[1] & alarm_entity.supported_features, ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ AttributeError: 'NoneType' object has no attribute 'supported_features'
[...]

PR #1408, opened 2026-04-19
Guard against None alarm_entity in ready_to_arm_modes_changed

optimusbasti:
## Problem

`EventHandler.async_handle_ready_to_arm_modes_changed` looks up the alarm entity for the given area and then accesses `alarm_entity.supported_features`. When the area is the master and no children are configured yet, or when the area entity has been removed at runtime, the lookup returns `None` and the subsequent attribute access raises, which kills the event handler.

## Fix

Add an early return with a debug log when `alarm_entity` is `None`. Same approach as the existing PR #1373 — posting as a separate PR for visibility while that one is idle.

## Behavior

- Normal setup with configured areas: unchanged.
- Master area with no children: event is silently ignored and a debug log line is emitted instead of a traceback.
- Entity removed mid-session: same.

## References

- References #1373

### DIFF
--- a/custom_components/alarmo/event.py
+++ b/custom_components/alarmo/event.py
@@ -33,11 +33,10 @@ class EventHandler:
             alarm_entity = self.hass.data[const.DOMAIN]["master"]
 
         if alarm_entity is None:
-            _LOGGER.debug(
-                "Ignoring %s for area %s: alarm entity is None "
-                "(likely removed or master not configured)",
-                event,
+            _LOGGER.warning(
+                "Cannot resolve alarm_entity for area_id: %s (event: %s)",
                 area_id,
+                event,
             )
             return
 
--- a/custom_components/alarmo/event.py
+++ b/custom_components/alarmo/event.py
@@ -32,6 +32,15 @@ class EventHandler:
         else:
             alarm_entity = self.hass.data[const.DOMAIN]["master"]
 
+        if alarm_entity is None:
+            _LOGGER.debug(
+                "Ignoring %s for area %s: alarm entity is None "
+                "(likely removed or master not configured)",
+                event,
+                area_id,
+            )
+            return
+
         if event in [
             const.EVENT_FAILED_TO_ARM,
             const.EVENT_COMMAND_NOT_ALLOWED,
@@ -75,12 +84,6 @@ class EventHandler:
             self.hass.bus.async_fire("alarmo_command_success", data)
 
         elif event == const.EVENT_READY_TO_ARM_MODES_CHANGED:
-            if alarm_entity is None:
-                _LOGGER.debug(
-                    "Ignoring ready_to_arm_modes_changed for area %s: alarm entity is None (likely removed or master not configured)",
-                    area_id,
-                )
-                return
             supported_modes = dict(
                 filter(
                     lambda el: el[1] & alarm_entity.supported_features,
--- a/custom_components/alarmo/event.py
+++ b/custom_components/alarmo/event.py
@@ -1,9 +1,13 @@
 """fire events in HA for use with automations."""
 
+import logging
+
 from homeassistant.core import callback
 from homeassistant.helpers.dispatcher import async_dispatcher_connect
 
 from . import const
+
+_LOGGER = logging.getLogger(__name__)
 
 
 class EventHandler:
@@ -71,6 +75,12 @@ class EventHandler:
             self.hass.bus.async_fire("alarmo_command_success", data)
 
         elif event == const.EVENT_READY_TO_ARM_MODES_CHANGED:
+            if alarm_entity is None:
+                _LOGGER.debug(
+                    "Ignoring ready_to_arm_modes_changed for area %s: alarm entity is None (likely removed or master not configured)",
+                    area_id,
+                )
+                return
             supported_modes = dict(
                 filter(
                     lambda el: el[1] & alarm_entity.supported_features,


custom_components/alarmo/event.py
```diff
@@ -71,6 +75,12 @@ def async_handle_event(self, event: str, area_id: str, args: dict = {}):
             self.hass.bus.async_fire("alarmo_command_success", data)
 
         elif event == const.EVENT_READY_TO_ARM_MODES_CHANGED:
+            if alarm_entity is None:
```
nico97118: I agree it's a good catch to check the alarm entity was resolved.
However, I don't understand why you inserted this check at this point.

I'm feeling this would be better suited at line 33 Just after the resolving snippet : 
```python
        if area_id:
            alarm_entity = self.hass.data[const.DOMAIN]["areas"][area_id]
        else:
            alarm_entity = self.hass.data[const.DOMAIN]["master"]
```

From my understanding, all the code after this point doesn't really make sense if the alarm_entity is "None"

nico97118 (commented):
Hello,

This is a good catch, it's a good practice to check alarm_entity was resolved before doing anything in `async_handle_event`

However, I'm just wondering why adding the check that late in the code (see comment)

optimusbasti:
Good point, moved the check to right after the `alarm_entity` resolution in 1ae0207. The other branches (`EVENT_FAILED_TO_ARM`, `EVENT_ARM`, etc.) already dereference `alarm_entity.entity_id` too, so the early-return benefits all code paths, not just `ready_to_arm_modes_changed`. Thanks for the review!

nico97118 (approved):
Thank you for considering my review.
I have no comment left :)

optimusbasti:
hi @nielsfaber, friendly bump — @nico97118 reviewed and approved, CI is unstable but mergeable. would appreciate a merge when you get a moment. thanks!

optimusbasti:
Hi @nielsfaber, second friendly bump on this one. Approved by @nico97118 on April 20, mergeable, and the CI flake is unrelated to the change. Anything still outstanding from your side, or is this ready for the next release window? Happy to rebase if needed.

nielsfaber:
@optimusbasti 
Thank you for your contribution.
Please do not post bumping messages in the future, it is completely unnecessary.

I'm having some trouble understanding this PR.
It seems to mask the underlying issue rather than adressing it.
You state the following:
> EventHandler.async_handle_ready_to_arm_modes_changed looks up the alarm entity for the given area and then accesses alarm_entity.supported_features. When the area is the master and no children are configured yet, or when the area entity has been removed at runtime, the lookup returns None and the subsequent attribute access raises, which kills the event handler.

I would expect the lookup to never result into None.
The event is always fired by an alarm entity.
In the code this event is fired by either the area entity
https://github.com/nielsfaber/alarmo/blob/1c1f097a3fe87fe8548bbf34a7ed7048b8fbf737/custom_components/alarmo/alarm_control_panel.py#L1192-L1198

Or by the master entity
https://github.com/nielsfaber/alarmo/blob/1c1f097a3fe87fe8548bbf34a7ed7048b8fbf737/custom_components/alarmo/alarm_control_panel.py#L1587-L1593

You state that
> When the area is the master and no children are configured yet

This situation should not occur.
Master alarm is only active if at least 2 areas are configured and the user has enabled the master entity.

If it does occur, it is likely an artifact of changed configuration and the code doesn't properly handle cleaning up the master alarm which is no longer active.

Also you state that the issue happens when
> the area entity has been removed at runtime

This seems to indicate the same behaviour.
The code should clean up an obsolete entity and the ready_to_arm_modes_changed should never be fired by it.

In summary, I think this PR adresses the issue in the wrong way. I think that instead we should try to find the actual cause of the issue (presumably the incomplete cleaning up of entities no longer in use) and eliminate the situation.

Do you have managed to reproduce the issue?
If so, can you explain a bit what steps are needed to run into this?
Hopefully we can we trace back the actual problem from here.

nico97118:
>It seems to mask the underlying issue rather than adressing it.

@nielsfaber while I agree with you that this PR will not address the root cause of the issue experienced by @optimusbasti. I'm feeling this safe check is still a good contribution.

It's always a good thing to protect your code by making sure a resolved variable result in something we can deal with.
Here, even though `alarm_entity` should always been resolved considering the algorithm logic : sometimes bugs happens.

If this PR doesn't figure out the root cause of the bug. It still prevents `async_handle_event(...)` to run with an uncontrolled behavior.

In this case, it might not be a big mistake since it result in a failure when trying to access a attributes of the entity.
However, it's always a good thing to prevent your functions to run with undesired values -> resulting in undefined behaviors. As a cyber-sec engineer, I can tell this is the kind of bugs an attacker would seek and try to abuse.

I'm not super worried about this case since the error is caught up soon enough. 
But it worth mentioning that checking `alarm_entity` is defined after resolving it from `area_id ` seems like a clean-code improvement here and I am aligned with this philosophy.

> I think that instead we should try to find the actual cause of the issue (presumably the incomplete cleaning up of entities no longer in use) and eliminate the situation.

According to me, we should do both. Keep investigating on the root cause, but still prevent async_event_handler to run with alarm_entity undefined.

**note**: it worth reconsidering the `debug` log to `error` level (`warning` at least) since this is not an normal behavior.
Also reword the log "Ignoring %s for area %s: alarm entity is None " to explain the error a bit more. (e.g. "Cannot resolve alarm_entity for area_id : %s" )

optimusbasti:
Apologies for the bumps @nielsfaber, won't happen again.

On the masking question — fair pushback. Reproduction context for the original crash:

**Setup:** Alarmo with a master area and 3 child areas. `automations.yaml` had three notification automations bound to `alarmo_failed_to_arm` events and another listening on master via the `area_id: master` flow.

**Steps to repro the original crash:**
1. Configure master + 3 child areas, save, restart HA.
2. Trigger an arm command on the master area while one child area was deleted between the modes-changed event firing and the handler running (race observable when the area was removed via the UI without restarting). The cached `hass.data[const.DOMAIN]["areas"][area_id]` entry returns `None` because the storage update replaces the dict but `async_handle_ready_to_arm_modes_changed` already had `area_id` in flight.
3. `alarm_entity.supported_features` raises `AttributeError` and the event handler dies, which leaves Alarmo in a state where the next arm command on master also fails because the bus listeners are torn down.

I had logs of three crashes from my own setup before I patched it locally, all on the master path. The race is small but reproducible by deleting an area while `alarmo_modes_changed` is firing.

Per your feedback I've also pushed (`commit pending`):
- Log level changed to `WARNING` (was `debug`) so the symptom isn't swallowed silently
- Message reworded to `Cannot resolve alarm_entity for area_id: %s (event: %s)` so it's actionable in HA logs

Agreed @nico97118's framing — this is a defence-in-depth guard that surfaces the problem rather than masking it. The WARN log gives enough signal to chase the root cause if someone hits this path again. Happy to dig further into the area-storage lifecycle if you want a follow-up PR for the actual race; that one needs more time to sit with.

nielsfaber:
Thanks for the response and the changes made.
Agreed that this is a good improvement that will be directly helpful for users who experience the issue.
If the underlying root cause of the issue can identified and adressed this would be my preferred long-term solution.
At the moment I don't have the time available to dig into this myself, so this can be picked up later by either me or you.

PR #1420, opened 2026-05-10
fix(auth): restore threading for bcrypt work from #1182

nico97118:
Following the feedback of @therealmrfox in #1394 

I assume there is obviously a win by manually threading the bcrypt work on some setup and configuration (e.g. many alarmo codes to check and a weak HW such as RPI)

while I still think it's better to offload this check via `hass.async_add_executor_job` since it prevent this work to block the event loop. It seems that it is still a good improvement to parallelize codes validations to speed up the code/user resolution.

I don't see any way to handle this with a cleaner way at this moment.

Since this code lived in Alarmo for a while without raising any issues : 
I propose to add it back so users still benefit from the parallelization added by @therealmrfox while we still avoid blocking the event loop, and we still avoid resolving the code multiple times when called on Alarmo's master entity.

### DIFF
--- a/custom_components/alarmo/__init__.py
+++ b/custom_components/alarmo/__init__.py
@@ -3,6 +3,7 @@
 import re
 import base64
 import logging
+import concurrent.futures
 
 import bcrypt
 from homeassistant.core import (
@@ -48,6 +49,8 @@ from .automations import AutomationHandler
 
 _LOGGER = logging.getLogger(__name__)
 
+# Max number of threads to start when checking user codes.
+MAX_WORKERS = 4
 # Number of rounds of hashing when computing user hashes.
 BCRYPT_NUM_ROUNDS = 10
 
@@ -369,11 +372,17 @@ class AlarmoCoordinator(DataUpdateCoordinator):
                 return check_user_code(self.store.async_get_user(user_id), code)
 
             users = self.store.async_get_users()
-            for user in users.values():
-                result = check_user_code(user, code)
-                if result:
-                    return result
-            return None
+            with concurrent.futures.ThreadPoolExecutor(
+                max_workers=MAX_WORKERS
+            ) as executor:
+                futures = [
+                    executor.submit(check_user_code, user, code)
+                    for user in users.values()
+                ]
+                for future in concurrent.futures.as_completed(futures):
+                    if future.result():
+                        executor.shutdown(wait=False, cancel_futures=True)
+                        return future.result()
 
         return await self.hass.async_add_executor_job(_authenticate_user_sync)
 


therealmrfox:
Thanks! I really appreciate the quick update!

> I don't see any way to handle this with a cleaner way at this moment.

I'm not a particular fan of the threading either. After thinking about the problem a little more, I think an argument can be made to further reduce the number of bcrypt rounds (instead of or in addition to this change). 

It boils down to the threat scenario the hashing is supposed to address:

- Attacker with user-level access to HA (e.g., the person who broke into my house and is trying to disable the pending alarm): Subject to the same hardware limitations, and adding a per-user lockout after invalid guesses is a better way to defend against this.
- Attacker with root access to HA: Can already do everything they could if they had the code.
- Attacker with offline access to HA data (e.g. access to unencrypted backup): This is the scenario where bcrypt helps - sometimes. 
  - For the typical 6-or-so-digit PINs most people use with an alarm system, it's game over anyway.
  - Truly random high-entropy codes (like a 128 bit RFID tag): Should be hard to brute force even with few bcrypt rounds (though of course extra rounds make it even more expensive).
  - Typical alphanumeric password: This is where I'd see the biggest benefit of extra bcrypt rounds to make dictionary attacks more expensive. I'm not sure how many people use guessable alphanumeric passwords for their Alarmo - probably not that many. OTOH, an additional risk here is that those passwords might be reused - so it could be "I found your HA backup and have your bank password", not just "I found your HA backup and can disable your alarm".

So, there's a tradeoff to me made. Ultimately @nielsfaber may want to think about this (and please double check my homework above - I may be missing something, and there are other factors to consider, e.g. the possibility of DoS'ing HA by starving it of CPU/threads with repeated "disarm" requests).

nico97118:
Hey @therealmrfox

I'm still thinking about this in the corner of my mind. I think I start to see a way to handle this, but it will require deep changes in Alarmo's password storage/authentication flow.

Food for thought: [optimizing-api-authentication-hmac](https://vesselapi.com/blog/optimizing-api-authentication-hmac)

Even though the topic is different, I think the underlying problem is actually very similar to ours:
authenticating a salted secret without having a public identifier associated with it.

In a traditional authentication flow using bcrypt, authentication is typically O(1):
you first identify the user, retrieve a single salted hash, then run one bcrypt.checkpw() against it.

In Alarmo's case, we only receive a code with no associated public identifier.
This means we currently have to test the submitted code against every stored bcrypt hash:
O(n_users * bcrypt.checkpw()).

The core issue is that bcrypt hashes are intentionally non-deterministic because of salting.
So we cannot simply hash the submitted code once and compare it against stored bcrypt hashes.

One possible solution would be to introduce a deterministic lookup key derived from the code itself, while still preserving bcrypt as the final password verifier.

What came to my mind (inspired by the article above) is using a peppered HMAC as a secure lookup index.

Something like:

lookup_key = HMAC(server_secret, password)
password_hash = bcrypt(password)

Stored as:

lookup_key | bcrypt_hash

Authentication flow would then become:

1. compute the lookup_key from the submitted code using the server secret
2. retrieve the corresponding bcrypt hash using that lookup_key
3. run a single bcrypt.checkpw() against the retrieved bcrypt hash

That would move us from:
`O(n_users * bcrypt.checkpw())`

to approximately:
`O(1 * HMAC + 1 * bcrypt.checkpw())`

The HMAC would only serve as a deterministic lookup mechanism.
bcrypt would still remain the actual password verifier and offline protection layer.

This is conceptually close to how peppering/keyed verifiers are described in OWASP/NIST recommendations.

----
One important clarification regarding the threat model: the lookup_key derived from HMAC is only a deterministic indexing mechanism and does not replace password verification. **If the server secret were ever compromised, an attacker could indeed brute-force the small PIN space and recover valid codes from the lookup table.**

However, bcrypt still plays a meaningful role in a different and independent scenario: it protects against offline attacks on the stored database alone. In that case, even with full DB access, an attacker cannot efficiently recover valid codes without paying the bcrypt cost per candidate. This preserves a defense-in-depth property, where HMAC optimizes lookup performance while bcrypt remains the actual slow-verification barrier against database compromise.

To sum it up, switching to this solution would create an additional attack surface.
- Currently, if DB leaks, an attacker has no choice but try to brute force through bcrypt
- If we use HMAC(secret, code) as lookup_key, an attacker who get his hand on secret AND the DB could easily retrieve pins from the lookup_key.

It worth considering an attacker who his able to retrieve both the DB AND the server secret has probably already won and has more power over your setup than simply breaks your pincode.

I'll leave it like that for now and wait for your thoughts (and leave me some time to think about it some more time)

therealmrfox:
I love the detailed analysis! 

Agree what makes this tricky is that we don't know which user is being authenticated. Similar to the article you linked. I also think the proposed lookup-index based approach is promising, though as described, comes with its own security considerations. 

It is also, obviously, a sizable change. Is it worth it? 

> If the server secret were ever compromised, an attacker could indeed brute-force the small PIN space and recover valid codes from the lookup table.

This is true for the current bcrypt approach too, if an attacker has access to the user table. With 6-8 digit PIN codes typically used for alarm systems, no amount of salting, peppering, hashing, or bcryp-ing will meaningfully delay an attack. E.g. ~3 mins with a single 4090 for an 8 digit PIN per [this](https://specopssoft.com/blog/hashing-algorithm-cracking-bcrypt-passwords/) (though they don't specify the # round they used unless I missed it). 

Another thing to consider (more of an implementation detail, and a bit tangential) is multiple users with the same password/PIN, who would obviously match to the same lookup key. Ideally this should not be allowed (it also causes problems if those users have different permissions). 

I attempted to enforce that in #1211, but the logic is incorrect because the read-modify-write of the user table isn't transactional. This is where I stumbled into the slowness from removing the parallelization - I added a new user, the check for existing users with same password ran slowly, and when nothing happened in the UI I hit "save" again - ending up with 2 users with the same name and pw.  So ironically #1211 might actually *increase* the likelihood of duplicate passwords, and I should probably revert it or think about proper locking/transactions.

nico97118:
Hi @therealmrfox 

> It is also, obviously, a sizable change. Is it worth it?

Hard question. My inner me would definitely goes for "It is worth it because it moves towards the right design.
But let's be honest, we are probably 3 to 4 users concerned by this 😆 

> This is true for the current bcrypt approach too, if an attacker has access to the user table. With 6-8 digit PIN codes typically used for alarm systems, no amount of salting, peppering, hashing, or bcryp-ing will meaningfully delay an attack. E.g. ~3 mins with a single 4090 for an 8 digit PIN per [this](https://specopssoft.com/blog/hashing-algorithm-cracking-bcrypt-passwords/) (though they don't specify the # round they used unless I missed it).

Considering your point - which is solid - it might worth considering what does bcrypt brings to the table.
Let me explain what's on my mind :

Bcrypt is slow on purpose, its usefulness is to voluntarily make brute-force CPU expensive. So an attacker who wants to brute force the password through a bcrypt.checkpw would be slowed down by the CPU load (CPUs of the server if the brute force is made online, of his setup for an offline attack).

In the meantime, we are trying to reduce the time of each try by spreading the cost. which somewhat defeats the original purpose of bcrypt
If the goal is to instantly find and authenticate a pincode. A keyed peppered hash stored would make a pincode validation way faster.

Let's imagine we switch the pincode/password storage to a simple peppered hash table.
The authentication flow would be : 
1. compute the peppered hash of submitted code
2. check if such a hash exists in the table
3. return true if yes, otherwise return false -and maybe run some protection against brute-force (e.g. 3 consecutive fails trigger an event, trigger the alarm or whatever)

On the security side, you are strongly protected against offline database compromise as long as the server secret remains uncompromised - an offline attack would be quite hard without the pepper. 
OTOH, an online attack could potentially get easier if the anti-brute force mechanism can be bypassed.

If both the database and the server secret are compromised, it's trivial, it's easier than with bcrypt, but as you said, bcrypt isn't that solid with a small space either (~3 mins with a single 4090 for an 8 digit PIN) 

According to this train of thought, a peppered keyed deterministic verifier might better fit our use case than the bcrypt flow since we are kinda overuse the bcrypt.pwcheck by running it against all existing salted hash until it match.

At the end of the day, the authentication flow could be a cheap, mostly instant,  `O(1 * HMAC(pepper, code))`.
While [OWASP discourage fast hash](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html) for password storage. It does not fit exactly a pincode verification with online protections.
Right now, we are using a user/password storage model for a pin verification.

Now I've said all of that:
Going that way would implies a drastic change of password management, introduce a server secret - needs to be configured/generated at setup and migrated with HA if you want to keep your hash table usable after migration.
It also means we got to deal with currently stored salted hash to avoid breaking users deployed setup.. meaning falling back on bcrypt.checkpw every salted hash if peppered hash isn't found - and maybe store the peppered hash if a salted hash is found via bcrypt in order to slowly migrate already existing code on users setup.

Really not sure @nielsfaber would be down to encourage such a tricky change.

-----

>Another thing to consider (more of an implementation detail, and a bit tangential) is multiple users with the same password/PIN, who would obviously match to the same lookup key. Ideally this should not be allowed (it also causes problems if those users have different permissions).

I must admit I've never checked the code regarding this topic but since pincode/password are both the identifier and the authentifier, having 2 "users" with the same code make no sense in my opinion and should definitely not be doable.

If 2 users are created with the same password, bcrypt will generate a different salt, so a different salted-hash. But whenever the duplicated code is submitted, it will always hit the first existing matching entry. There is no way to differentiate which user entered the code, the first one found in the table would be the one "authenticated"
On this I will add that checking - once again - if that user (code) exists would be easier via a peppered hash. But that's the whole point discussed above ahah

nielsfaber:
I don't really want to change the encryption method for the codes.
My knowledge about the subject is rather limited, I deliberately chose the same implementation as HA uses for validating login credentials (at least at the time) to achieve similar level of security.
I understand that it takes some cpu time for validating the code, but it should be insignificant (else your HA hardware is likely a bottleneck).
If we can optimize the code for speed improvement, that is totally fine.
But changing the authentication method is not. It would result in a rather nasty situation anyway, because we would need to support both the old/legacy and new authentication method, so duplicate code.
I expect at least 90% of users to have 1, maybe 2 pin codes. As such I don't think the whole subject is a priority.

@nico97118 Is this PR ready for merge? It's not completely clear to me given the discussion that followed.

nico97118:
Yep, this is mergeable.

I'm not a huge fan of this code but it obviously improve bcrypt work.

It is still compatible with my previous rework of the authentication flow. 
Previous work added better logic of the auth process and avoid bcrypt to block the event loop.

I expected the previous PR would have enhanced the situation enough to avoid keeping threading in alarmo.

As I said, this threading is literally going against bcrypt original purpose. But I got your point concerning changing the auth method and I agree such a change should not be made in a hurry.

PR #1435, opened 2026-06-18
Fix Override Code not working (Update alarm_control_panel.py)

HTS126:
EDIT: The documentation states that this is NOT what the function should return, but it happens to work perfectly normally when changed. It can be used as a temporary 'fix' until an actual fix is implemented. 

I can't promise that I haven't missed something that would inadvertently break something else, but all seems to work for me!

---

This is a quick fix I have found for correcting the issue in the latest release where codes marked as override codes would no longer force-arm the system, bypassing open sensors.

### DIFF
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -404,7 +404,7 @@ class AlarmoBaseEntity(AlarmControlPanelEntity, RestoreEntity):
 
         # success
         self._changed_by = user[ATTR_NAME]
-        return True, None
+        return True, user
 
     async def async_service_disarm_handler(self, code, context_id=None):
         """Handle external disarm request from alarmo.disarm service."""


nielsfaber:
Your proposed fix makes no sense to me.
The function documentation states:
```python
     async def _validate_code(self, code, to_state):  # noqa PLR0911
        """Validate code and user permissions for a requested state change.

        Returns a (success, error_event) tuple. When success is True,
        error_event is None.
        """
```
So the line you propose to change:
https://github.com/nielsfaber/alarmo/blob/4930747bdd79be3e7a83c85ae5d177f32c8b0eeb/custom_components/alarmo/alarm_control_panel.py#L407

Is correct as it is.
We should definitely not return `user` here, this parameter is reserved for `error_event`, which is only passed in case the first parameter is `False`.

HTS126:
(Edit: Rephrased to make it read better)

Apologies, I didn't see the documentation part originally, I just had a quick scroll through as I was trying to get it working quickly so my dad could keep using the alarm system. 

Saw the parameter called 'code' and thought it might be referencing the user's inputted code, and it happened to work as expected with no other obvious problems when I changed it (but I think that must just be a coincidence).

HTS126:
@nielsfaber  Just as a note, the function 
```
async def async_handle_arm_request(self, arm_mode, **kwargs): ...
```

contains 

```
(res, info) = await self._validate_code(code, arm_mode)
```

and then checks the override code with 

```
elif info and info[const.ATTR_IS_OVERRIDE_CODE]:
    bypass_open_sensors = True
```

But, if the documentation is correct that `_validate_code` returns a (success, error_event) tuple, then there is never a valid `info` parameter because it's just returning `None`. 

I'm guessing this is why my temporary fix worked as it now passes the user info as the 'info' parameter and it's able to check that value.

Issue #1430, opened 2026-05-25
Override code no longer force-arms / bypasses open sensors in Alarmo 1.10.18


dannybloe:
### Checklist

- [x] I am using the latest version of Alarmo (latest version can be found [here](https://github.com/nielsfaber/alarmo/releases/latest))
- [x] I checked for similar existing requests (both open and closed) before posting.

### Alarmo Version

1.10.18

### HA Version

2026.5.4

## Version

- Alarmo firmware: `1.10.18`
- Home Assistant version: `2026.5.4`
- Alarmo card: `custom:alarmo-card`
- Additional input method: Z-Wave keypads via AppDaemon service call

## Expected behavior

Normal codes should behave normally:

```text
normal code -> arm normally, fail if sensors are open
```

Override codes should bypass open sensors:

```text
override code with "Is override code: true" -> arm and bypass open sensors
```

## Actual behavior

A code configured with **"Is override code: true"** behaves the same as a normal code.

When sensors are open, Alarmo refuses to arm and fires `alarmo_failed_to_arm` with:

```yaml
reason: open_sensors
```

## Confirmed working

Calling `alarmo.arm` manually from **Developer Tools -> Actions** with the same code and `force: true` works:

```yaml
entity_id: alarm_control_panel.alarmo
mode: away
code: "<override-code>"
force: true
```

This successfully arms the alarm and bypasses the open sensor.

Also, in `custom:alarmo-card`, using a normal code first shows the open sensors warning, and pressing the **Bypass sensors** button successfully arms the alarm.

## Confirmed failing

Using the same override code without explicitly passing `force: true` fails.

This happens from multiple clients/input paths:

- `custom:alarmo-card`
- Z-Wave keypads sending the same code to Alarmo via an internal service call from an AppDaemon script

So this does not appear to be limited to the Alarmo card/frontend.

## Failing event

```yaml
event_type: alarmo_failed_to_arm
data:
  command: arm_away
  context_id: null
  area_id: "xxx"
  entity_id: alarm_control_panel.alarmo
  reason: open_sensors
  sensors:
    - binary_sensor.xxx
origin: LOCAL
time_fired: "2026-05-25T12:46:45.176664+00:00"
context:
  id: xxx
  parent_id: null
  user_id: null
```

### Steps to reproduce

1. Configure a code with **"Is override code: true"**.
2. Leave one protected sensor open.
3. Try to arm `alarm_control_panel.alarmo` using that override code.
[...]

PR #1444, opened 2026-07-05
Second attempt to fix override code not working

HTS126:
Managed to sit down and actually look at this properly (for Issue #1430). Previous attempt was PR #1435 but that was just a quick temporary fix so we could use the alarm and I missed the documentation of the function I changed.

The problem seems to be that the function 
```
async def async_handle_arm_request(self, arm_mode, **kwargs): ...
```

contains 

```
(res, info) = await self._validate_code(code, arm_mode)
```

and then checks the override code with 

```
elif info and info[const.ATTR_IS_OVERRIDE_CODE]:
    bypass_open_sensors = True
```

But the documentation says  that `_validate_code` returns a (success, error_event) tuple. So there is never a valid `info` parameter because it's just returning `None`. 

To fix this, I have added a new attribute `self._validated_user` which is initialised to None originally, but on a successful authentication, it stores the user that tried to arm the system. From this, the user info can be checked by the original caller function to see if it contains the override code attribute. 

Again, I have tried to test this the best I can and it seems to work correctly for me, but would appreciate some feedback from others!

### DIFF
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -176,7 +176,6 @@ class AlarmoBaseEntity(AlarmControlPanelEntity, RestoreEntity):
         self._revert_state = None
         self._ready_to_arm_modes = []
         self._last_triggered = None
-        self._validated_user = None
 
     @property
     def device_info(self) -> dict:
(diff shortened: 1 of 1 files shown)


nielsfaber:
You are absolutely right about this issue!
This was clearly an oversight when the code validation part was restructured.
Now that we know the culprit, I honestly think your previous PR was more elegant than introducing a new variable.
I think that if it is made clear what the second variable holds, it can be perfectly repurposed for returning user information.

So if we change:
>  Returns a (success, error_event) tuple.
When success is True, error_event is None.

Into something like:
> Returns a (success, info) tuple.
When success is True, info is the user data.
When success is False, info is the error event.

We could continue with your original approach.
I would prefer this over introducing a new variable, as this variable is only used in one place within the code.

If you would be so kind to re-open your previous PR, I'd be happy to merge it.

HTS126:
Unfortunately I closed the branch that the previous PR was tied to so I'm not able to reopen it, so I've made the changes to the code linked to this PR. 

Changes: 
- Reverted the previous attempt to fix and reimplemented my first fix.
- Changed the documentation of the _validate_code() function to show it returns a (success, info) tuple and what each of those means in different states.

nielsfaber:
Thanks a lot for the extra efforts, it's much appreciated!
Let's merge your fix, I will include it in the next release.
Note that next release may take a while, since I try to bundle multiple changes.

Issue #845, opened 2023-11-22
 "Allow Open Initially" sensor names in Armed alert as a wildcard same as "Bypassed"


ideal2545:
### Checklist
- [ ] I checked for similar existing requests (both open and closed) before posting.
- [ ] My request is generic, other users may benefit from it too.

### Proposal
Right now, Alarmo allows us to use the Bypassed wild card when arming the alarm to see what is bypassed, I think it would be great to have the same data for Open Initially. 

This request is so that users can see what windows are open so that they can keep track and close the windows later that are open. 

### Additional info
[some example of how the new functionality should look like]

nielsfaber:
I agree with your request.
I just don't really know (yet) how to make this happen in the code. 
The bypassed sensors are handled differently than sensors which are initially ignored. So I might need to reshuffle some things to be able to add this feature.

zd3sf:
Okay so I did some digging because I had a similar request open. Here are my findings:

1) Open sensors get registed inside `open_sensors` attribute
2) If alarm is forced and there are open sensors, those open sensors get registered under `bypassed_sensors`
3) If sensors are `allow open initially`, then they're not registered under either `open_sensors` or `bypassed_sensors`

here's the append code that handles it
![image](https://github.com/user-attachments/assets/fc56362f-accb-44d1-bc62-6191b505d761)

The only solution is to create a new attribute, I called mine `allowed_open`  but it needs some rewrite:

### In Sensors.py
![image](https://github.com/user-attachments/assets/02938299-7ef4-469a-ae3b-19325d85084c)
![image](https://github.com/user-attachments/assets/3cc94f8b-3f35-4f64-b7ee-330474c72d9e)

### In alarm control panel.py
![image](https://github.com/user-attachments/assets/3d28f051-a9cb-4501-ad4b-2bc58816f3dc)
![image](https://github.com/user-attachments/assets/8daa5709-a7bc-46c5-ba03-694618b27e39)
![image](https://github.com/user-attachments/assets/5bb06965-317b-43d6-b350-e463ea23b84c)
![image](https://github.com/user-attachments/assets/49a5ec16-249f-4981-8d45-10d0a62513e9)
![image](https://github.com/user-attachments/assets/f1c805bc-b80c-46f1-913a-fa4739475c8c)
![image](https://github.com/user-attachments/assets/fc694a47-5ec2-4868-8fd0-1ba9e4aaf9ed)

###In automations.py
[...]

PR #1445, opened 2026-07-10
Add attribute "active_when_armed" which can be used as a wildcard for notifications

zd3sf:
An atrribute added to Alarmo control panel that records the sensors that are active during an arm event. This would function similarly to the attribute bypassed_sensor. Works for sensors set as "allow open initially" in Alarmo sensor settings. Solves https://github.com/nielsfaber/alarmo/issues/845 

image

image

image

### DIFF
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -270,6 +270,24 @@ class AlarmoBaseEntity(AlarmControlPanelEntity, RestoreEntity):
         else:
             self._open_sensors = {}
 
+##########NEW ADDITION
+    @property
+    def active_when_armed(self):
+        """Get allowed open."""
+        if not self._active_when_armed:
+            return None
+        else:
+            return self._active_when_armed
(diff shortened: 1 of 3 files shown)


nielsfaber:
Thank you for contributing to Alarmo.
I disagree on intoducing a separate attribute for sensors that are open during arming.
For this the `open_sensors` attribute is intended.
What are your reasons for adding a new attribute for this scenario?

zd3sf:
Hi

If alarm is forced and there are open sensors, those open sensors get registered under bypassed_sensors
If sensors are allow open initially, then they're not registered under either open_sensors or bypassed_sensors, hence the new attribute and wildcard.

nielsfaber:
> If sensors are allow open initially, then they're not registered under either open_sensors or bypassed_sensors, hence the new attribute and wildcard.

My suggestion is to change your PR to have these sensors listed under open_sensors.
I don't see the need or benefit for introducing a new attribute, I think it will only raise confusion.

PR #1453, opened 2026-07-23
fix: skip allow_open sensors on startup evaluation to prevent false triggers

dennismdejong:
## Description

When Home Assistant restarts and sensors configured with "Allow open initially" (`allow_open: True`) are still open, Alarmo triggers the alarm unnecessarily. These sensors were open before the restart and should be allowed to remain open without triggering.

## Fix

In `_async_evaluate_armed_state_on_startup()` (`sensors.py`), added a check to skip sensors with `allow_open=True` that are currently open. This matches the existing behavior in `validate_arming_event()` which already handles `allow_open` correctly during normal arming.

## Related Issue

Fixes #1419

PR #1454, opened 2026-07-23
fix: do not ignore sensor transitions from unknown to open

### LINKED ISSUE #1450: Fast open/close sensor does not trigger alarmo when state goes from unknown to on
### Checklist

### Alarmo Version

v1.10.18

### HA Version

2026.7.2

### Bug description

While armed a door contact sensor was opened and closed within ~2 seconds. Alarmo did not transition to pending/triggered for this event at all - no state change was recorded for the alarm_control_panel entity.

About 6 minutes later, the same door was opened again, this time for only ~1 second, and this time Alarmo triggered normally (pending → triggered).

Both events are effectively identical in nature (a very brief door open/close while armed), yet only the second one caused Alarmo to react. This rules out sensor activation duration as the relevant factor - since the second, successful trigger was actually shorter (1s) than the first, failed one (2s).

- 05:51:11 — door sensor opened
- 05:51:13 — door sensor closed (~2s open)
→ No corresponding state change on alarm_control_panel.alarmo (stayed armed_away, no pending event)
- 05:58:20 — door sensor opened again
- 05:58:21(approx.) — door sensor closed (~1s open)
→ Alarmo changed to pending at 05:58:20
- 05:58:35 — Alarmo changed to triggered
- 05:59:49 — Alarmo disarmed

Image 

### Steps to reproduce

1. Arm Alarmo (armed_away) with a door contact sensor configured normally (no custom Delay on / Entry delay overrides).
2. Open and close the door briefly (1-2 seconds) — sometimes Alarmo triggers pending, sometimes it doesn't, with no clear pattern related to duration.

### Relevant log output

```shell

```

dennismdejong:
## Description

When a door contact sensor transitions UNKNOWN→OPEN→CLOSED (e.g., after temporary disconnection or restart), Alarmo treated the UNKNOWN→OPEN transition as "initial state" and ignored it. If the sensor then closes quickly, the open event is permanently lost.

## Fix

In `async_sensor_state_changed()` (`sensors.py`), replaced the blanket early-return on `old_state == STATE_UNKNOWN` with a state-aware check. Transitions from UNKNOWN to safe states (CLOSED/OFF) are still treated as initial. Transitions from UNKNOWN to violation states (OPEN) now proceed through normal trigger evaluation.

## Related Issue

Fixes #1450

### DIFF
--- a/custom_components/alarmo/sensors.py
+++ b/custom_components/alarmo/sensors.py
@@ -377,15 +377,28 @@ class SensorHandler:
         new_state = parse_sensor_state(event.data["new_state"])
         sensor_config = self._config[entity]
         if old_state == STATE_UNKNOWN:
-            # sensor is unknown at startup,
-            #   state which comes after is considered as initial state
-            _LOGGER.debug(
-                "Initial state for %s is %s",
-                entity,
-                new_state,
-            )
-            self.update_ready_to_arm_status(sensor_config["area"])
-            return
+            if new_state not in (STATE_OPEN, STATE_UNAVAILABLE) or (
+                sensor_config[ATTR_ALLOW_OPEN] and new_state == STATE_OPEN
+            ):
+                # transition to a safe state, or to open while the sensor is
+                #   allowed to be open — treat as initial state
+                _LOGGER.debug(
+                    "Initial state for %s is %s",
+                    entity,
+                    new_state,
+                )
+                self.update_ready_to_arm_status(sensor_config["area"])
+                return
+            else:
+                # transition to a violation state — do not treat as initial,
+                #   proceed through normal trigger evaluation
+                _LOGGER.debug(
+                    "Sensor %s recovered from unknown to %s while alarm is %s, "
+                    "evaluating as live state change",
+                    entity,
+                    new_state,
+                    self.hass.data[const.DOMAIN]["areas"][sensor_config["area"]].state,
+                )
         if old_state == new_state:
             # not a state change - ignore
             return
--- a/tests/test_startup_sensor_evaluation.py
+++ b/tests/test_startup_sensor_evaluation.py
@@ -427,3 +427,335 @@ async def test_sensor_closed_on_startup_no_trigger(
         await hass.async_block_till_done()
         assert_alarm_state(hass, ALARM_ENTITY, "disarmed")
         await cleanup_timers(hass)
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


nielsfaber:
The transition from unknown -> open is indeed filtered out by Alarmo.
However this is an intentional choice.
I expect a (binary) sensor to only have unknown state shortly after HA was restarted (otherwise I consider it a poor implementation of the integration).
The state change unknown -> open is most likely a recovery of the state prior to the restarting.

For example, consider a battery-operated contact sensor mounted on a window.
If the user chose to arm the alarm with the window open, the alarm should not be triggered after a restart.
Hence in this case
>  Transitions from UNKNOWN to violation states (OPEN) now proceed through normal trigger evaluation.

Is undesired behavior.
Perhaps your fix https://github.com/nielsfaber/alarmo/pull/1453 addresses this scenario?

nielsfaber:
@dennismdejong 
Could you respond to my feedback sent earlier?
I would like to prepare a next release for Alarmo, would be great if this PR can be merged before.

dennismdejong:
Thank you for the explanation — you're right, that scenario matters.

I've updated the fix: UNKNOWN→OPEN is now treated as an initial state when the sensor has `allow_open=True` ("Allow open initially"). So your example holds: if the user armed with the window open, the alarm will NOT trigger after a restart when the sensor recovers UNKNOWN→OPEN.

The behavior is now:

- UNKNOWN → safe state (closed/off): initial state, no trigger (unchanged)
- UNKNOWN → OPEN with `allow_open=True`: initial state, no trigger (new)
- UNKNOWN → OPEN with `allow_open=False`: processed as live state change, triggers normally (the #1450 race condition fix)

A test was added for the allow_open case (`test_sensor_unknown_to_open_with_allow_open_does_not_trigger`). PR #1453 addresses the related `_async_evaluate_armed_state_on_startup` path separately, and this update closes the gap for the state-change path too.

Branch force-pushed with the amended commit.

nielsfaber:
Thank you for the modified implementation. 
Given that the allow_open option is typically used for motion sensors, I think the filtered unknown->open transition, typically occurring at HA restart, is a good in-between solution.
Code looks good, your work can be merged 👍 .

nielsfaber:
@dennismdejong Can you please take care of the merge conflicts so this can be merged without issues?

nielsfaber:
Thank you!

PR #1455, opened 2026-07-23
fix: prevent duplicate notification execution with master entity

### LINKED ISSUE #1366: Notification Action for master performed multiple times
### Checklist

### Alarmo Version

V2.10.15

### HA Version

2026.1.3

### Bug description

I defined a notification action for master:

Image
I get this notification multiple times (used to get it only once). In my case 4 times (master plus 3 areas)

### Steps to reproduce

See description 

### Relevant log output

```shell

```

dennismdejong:
## Description

When a notification action is configured for the master entity, it fires multiple times per event (once per area + once for master). The master entity propagates commands to all areas, and each area's state change triggers the notification separately.

## Fix

In `async_alarm_state_changed()` (`automations.py`), when processing an area-level state change, skip execution if a master exists and the automation also has a master/global trigger — the master dispatch will handle it.

## Related Issue

Fixes #1366

### DIFF
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -323,9 +323,10 @@ class AlarmoBaseEntity(AlarmControlPanelEntity, RestoreEntity):
     async def _validate_code(self, code, to_state):  # noqa PLR0911
         """Validate code and user permissions for a requested state change.
 
-        Returns a (success, info) tuple. 
+        Returns a (success, info) tuple.
         When validation is successful, success is True, otherwise False.
-        When success is True, info is the user data (or None if no user/code was needed).
+        When success is True, info is the user data
+            (or None if no user/code was needed).
         When success is False, info is the error event.
         """
         # check bypass rules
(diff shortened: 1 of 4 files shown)


nielsfaber:
Could you explain your work?
I don't understand the reason behind the nested for loop you added.
To my understanding `validate_area` should handle that a notification configured for master entity will only be executed once the master changes.
Perhaps an extension to this function is required, rather than the added loop?

dennismdejong:
Good point — the nested loop was the wrong layer. I've moved the logic into `validate_area` instead.

The updated approach: when `area_id` is not `None` (individual area dispatch) AND a master entity exists, `validate_area` now returns `False` for triggers with an empty/`None` area. This keeps the master/global trigger from matching individual area dispatches — only master dispatches (`area_id=None`) will match.

The single additional line in `validate_area`:
```python
elif area_id and hass.data[const.DOMAIN].get("master"):
    return False
```

Tests still pass (99/99), branch force-pushed with the cleaner change.

nielsfaber:
Thank you for taking care of the changes, all looks good now 👍

PR #1456, opened 2026-07-23
fix: handle ALARMO_DISARM push action when arm_mode is None

### LINKED ISSUE #1447: Event ALARMO_DISARM doesn't work on Environment sensors
### Checklist

### Alarmo Version

v1.10.18

### HA Version

2026.7.2

### Bug description

it seems as if ALARMO_DISARM is not executed if alarmo was not armed (which is the case when an Environment sensor triggers)

### Steps to reproduce

The following works:
Alarmo is armed away/armed home
Environment sensor triggers Alarm
Event ALARMO_DISARM (triggered through an android notification) is triggered
Alarmo is disarmed

the following does not:
Alarmo is disarmed
Environment sensor triggers Alarm
Event ALARMO_DISARM (triggered through an android notification) is triggered
Alarmo is still triggered although it should be disarmed

### Relevant log output

```shell

```

dennismdejong:
## Description

When an environment sensor (always_on) triggers the alarm while disarmed, `arm_mode` is set to `None`. The push event handler validated `arm_mode` before checking the action type, causing `ALARMO_DISARM` to be rejected with "Cannot process the push action, since the arm mode is not known."

## Fix

In `async_handle_push_event()` (`__init__.py`), moved the DISARM action check before `arm_mode` validation with an early return. All other arm actions continue to validate `arm_mode` as before.

## Related Issue

Fixes #1447

### DIFF
--- a/custom_components/alarmo/__init__.py
+++ b/custom_components/alarmo/__init__.py
@@ -430,6 +430,11 @@ class AlarmoCoordinator(DataUpdateCoordinator):
                 )
                 return
 
+            if action == const.EVENT_ACTION_DISARM:
+                _LOGGER.info("Received request for disarming")
+                await alarm_entity.async_alarm_disarm(None, skip_code=True)
+                return
+
             arm_mode = (
                 alarm_entity._revert_state
                 if alarm_entity._revert_state in const.ARM_MODES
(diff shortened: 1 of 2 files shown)


nielsfaber:
Thank you for taking care of this.
Looks like a good fix to me 👍

Issue #1375, opened 2026-02-25
Alarmo not starting / no entity anymore


jpawlowski:
### Checklist

- [x] I am using the latest version of Alarmo (latest version can be found [here](https://github.com/nielsfaber/alarmo/releases/latest))
- [x] I checked for similar existing requests (both open and closed) before posting.

### Alarmo Version

v1.10.15

### HA Version

2026.2.3

### Bug description

Alarmo no longer starts and no entity is being created.

### Steps to reproduce

There were no changes besides HA updates for the last 2 months.
I guess what needs to be done to reproduce this is to update HA to its newest version.

### Relevant log output

```shell
Logger: homeassistant.components.alarm_control_panel
Quelle: helpers/entity_platform.py:680
Integration: Alarmzentrale (Dokumentation, Probleme)
Erstmals aufgetreten: 21:48:43 (1 Vorkommnis)
Zuletzt protokolliert: 21:48:43

Error adding entity alarm_control_panel.alarmo for domain alarm_control_panel with platform alarmo
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 680, in _async_add_entities
    await self._async_add_entity(
        entity, False, entity_registry, config_subentry_id
    )
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 998, in _async_add_entity
    await entity.add_to_platform_finish()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1393, in add_to_platform_finish
    await self.async_added_to_hass()
  File "/config/custom_components/alarmo/alarm_control_panel.py", line 779, in async_added_to_hass
    self.async_arm(self.arm_mode)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/config/custom_components/alarmo/alarm_control_panel.py", line 932, in async_arm
    ].validate_arming_event(
      ~~~~~~~~~~~~~~~~~~~~~^
        area_id=self.area_id,
        ^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        bypass_open_sensors=self._bypass_open_sensors,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/config/custom_components/alarmo/sensors.py", line 300, in validate_arming_event
    sensors_list = self.active_sensors_for_alarm_state(area_id, target_state)
[...]

Issue #1370, opened 2026-02-13
Entity no more available since new HA update


maathiieus:
### Checklist

- [x] I am using the latest version of Alarmo (latest version can be found [here](https://github.com/nielsfaber/alarmo/releases/latest))
- [x] I checked for similar existing requests (both open and closed) before posting.

### Alarmo Version

1.10.15

### HA Version

2026.2.1

### Bug description

Since i've updated my HA instance

### Steps to reproduce

Just using my normal script or button i can see the entity no more avaible

### Relevant log output

```shell
Enregistreur: homeassistant.components.alarm_control_panel
Source: helpers/entity_platform.py:680
intégration: Centrale d’alarme (documentation, problèmes)
S'est produit pour la première fois: 14:49:14 (1 occurrence)
Dernier enregistrement: 14:49:14

Error adding entity alarm_control_panel.alarmo for domain alarm_control_panel with platform alarmo
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 680, in _async_add_entities
    await self._async_add_entity(
        entity, False, entity_registry, config_subentry_id
    )
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 998, in _async_add_entity
    await entity.add_to_platform_finish()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1393, in add_to_platform_finish
    await self.async_added_to_hass()
  File "/config/custom_components/alarmo/alarm_control_panel.py", line 779, in async_added_to_hass
    self.async_arm(self.arm_mode)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/config/custom_components/alarmo/alarm_control_panel.py", line 932, in async_arm
    ].validate_arming_event(
      ~~~~~~~~~~~~~~~~~~~~~^
        area_id=self.area_id,
        ^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        bypass_open_sensors=self._bypass_open_sensors,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/config/custom_components/alarmo/sensors.py", line 300, in validate_arming_event
    sensors_list = self.active_sensors_for_alarm_state(area_id, target_state)
  File "/config/custom_components/alarmo/sensors.py", line 272, in active_sensors_for_alarm_state
[...]

Issue #1358, opened 2026-01-18
Error adding entity alarm_control_panel.alarmo for domain alarm_control_panel with platform alarmo


zylan1:
### Checklist

- [x] I am using the latest version of Alarmo (latest version can be found [here](https://github.com/nielsfaber/alarmo/releases/latest))
- [x] I checked for similar existing requests (both open and closed) before posting.

### Alarmo Version

1.10.15

### HA Version

2026.1.2

### Bug description

I can see this is associated with previous issue #1279 

I had alarmo configured and working for >18 months until one fateful update that broke it.

I've tried uninstalling Alarmo, deleting the entities, restarting, and using Spook's orphan cleanup tool, but I can't seem to progress past this error.

I couldn't reopen the issue, and wasn't sure where else to start?

Thanks

### Steps to reproduce

Updated to home assistant 2026.1, Alarmo 1.10.14

### Relevant log output

```shell
Logger: homeassistant.components.alarm_control_panel
Source: helpers/entity_platform.py:684
integration: Alarm control panel (documentation, issues)
First occurred: 7:51:12 AM (1 occurrence)
Last logged: 7:51:12 AM

Error adding entity alarm_control_panel.alarmo for domain alarm_control_panel with platform alarmo
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 684, in _async_add_entities
    await self._async_add_entity(
        entity, False, entity_registry, config_subentry_id
    )
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 1010, in _async_add_entity
    await entity.add_to_platform_finish()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1378, in add_to_platform_finish
    await self.async_added_to_hass()
  File "/config/custom_components/alarmo/alarm_control_panel.py", line 779, in async_added_to_hass
    self.async_arm(self.arm_mode)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/config/custom_components/alarmo/alarm_control_panel.py", line 932, in async_arm
    ].validate_arming_event(
      ~~~~~~~~~~~~~~~~~~~~~^
        area_id=self.area_id,
        ^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        bypass_open_sensors=self._bypass_open_sensors,
[...]

PR #1459, opened 2026-08-03
fix: keep sensor config as dict before HA startup completes

### LINKED ISSUE #1429: This entity is no longer being provided by the alarmo integration
### Checklist

### Alarmo Version

v1.10.18

### HA Version

2026.5.4

### Bug description

After updating Alarmo (I think), the alarm dashboard panel always displays 'This entity is no longer being provided by the alarmo integration. If the entity is no longer in use, delete it in settings.'.

I tried reinstalling Alarmo from HACS.

Image

### Steps to reproduce

1. Try Alarmo

### Relevant log output

```shell

```

vojtech-cerveny:
## Description

On a cold boot, Alarmo can fail to create its alarm entity entirely. The panel then shows *"This entity is no longer being provided by the alarmo integration"*, and the alarm is silently gone until the config entry is reloaded.

```
Error adding entity alarm_control_panel.alarmo for domain alarm_control_panel with platform alarmo
Traceback (most recent call last):
  ...
  File "/config/custom_components/alarmo/alarm_control_panel.py", line 819, in async_added_to_hass
    self.async_arm(self.arm_mode)
  File "/config/custom_components/alarmo/alarm_control_panel.py", line 972, in async_arm
    ].validate_arming_event(
  File "/config/custom_components/alarmo/sensors.py", line 303, in validate_arming_event
    sensors_list = self.active_sensors_for_alarm_state(area_id, target_state)
  File "/config/custom_components/alarmo/sensors.py", line 275, in active_sensors_for_alarm_state
    for entity, config in self._config.items():
AttributeError: 'NoneType' object has no attribute 'items'
```

## Root cause

`SensorHandler.__init__` defers `_setup_sensor_listeners()` — and therefore the first `_async_update_sensor_config()` that populates `self._config` — until `EVENT_HOMEASSISTANT_STARTED`:

```python
if hass.state == CoreState.running:
    ...
else:
    hass.bus.async_listen_once(EVENT_HOMEASSISTANT_STARTED, handle_startup)
```

But `AlarmoBaseEntity.async_added_to_hass()` restores the persisted state during *platform setup*, which happens earlier in the startup sequence. If the restored state is `arming`, it calls `async_arm()`, which reaches `validate_arming_event()` → `active_sensors_for_alarm_state()` and iterates `self._config` while it is still `None`.

In practice this means: restart HA during the exit delay, and the alarm entity is gone on the way back up.

This only reproduces on a cold boot (`hass.state != CoreState.running`), which is why reloading the config entry works around it — on reload the config is already in memory.

Note the restored `pending` / `triggered` states do **not** hit this: they call `async_trigger()`, which reads the area config rather than the sensor config.

## Fix

Initialize `_config` to `{}` and coerce a `None` from `store.async_get_sensors()` to `{}`.

"No sensors known yet" is the correct semantics at this point rather than a workaround: `_setup_sensor_listeners()` already re-evaluates every armed area through `_async_evaluate_armed_state_on_startup()` once HA has started, so the authoritative sensor evaluation still happens — just a moment later, as designed.

## Tests

Added `tests/test_startup_sensor_config_race.py`, which sets `CoreState.not_running` before setup to reproduce the real startup ordering, restores `arming`, and asserts the entity is added. Verified both ways against this branch:

- with `self._config = None` restored → fails with the exact `AttributeError` above
- with the fix → passes

Full suite: 104 passed. `ruff check` and `ruff format` clean.

Deliberately placed in a new file rather than appended to `tests/test_startup_sensor_evaluation.py` so that it does not conflict with #1454, which appends to that file. The two changes are orthogonal — #1454 fixes `async_sensor_state_changed()` for `UNKNOWN → OPEN` transitions at runtime.

## Related issues

Fixes #1429. The same traceback also appears in #1358, #1370 and #1375, all closed without a fix having landed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom_components/alarmo/sensors.py
+++ b/custom_components/alarmo/sensors.py
@@ -130,7 +130,11 @@ class SensorHandler:
 
     def __init__(self, hass: HomeAssistant):
         """Initialize the sensor handler."""
-        self._config = None
+        # The sensor config is only loaded once HA has finished starting (see
+        # _setup_sensor_listeners below). Until then this must be an empty dict
+        # rather than None: restoring a persisted alarm state can call into
+        # active_sensors_for_alarm_state() before that point.
+        self._config = {}
         self.hass = hass
         self._state_listener = None
         self._subscriptions = []
@@ -144,9 +148,10 @@ class SensorHandler:
         @callback
         def async_update_sensor_config():
             """Sensor config updated, reload the configuration."""
-            self._config = self.hass.data[const.DOMAIN][
-                "coordinator"
-            ].store.async_get_sensors()
+            self._config = (
+                self.hass.data[const.DOMAIN]["coordinator"].store.async_get_sensors()
+                or {}
+            )
             self._groups = self.hass.data[const.DOMAIN][
                 "coordinator"
             ].store.async_get_sensor_groups()
--- a/tests/test_startup_sensor_config_race.py
+++ b/tests/test_startup_sensor_config_race.py
@@ -1,0 +1,93 @@
+"""Test that Alarmo sets up before the sensor config has been loaded.
+
+SensorHandler defers loading its sensor config until EVENT_HOMEASSISTANT_STARTED,
+but the alarm entities restore their persisted state during platform setup, which
+happens earlier in the startup sequence. Restoring the 'arming' state calls
+async_arm(), which reaches into SensorHandler before its config is available.
+
+Regression test for https://github.com/nielsfaber/alarmo/issues/1429
+"""
+
+from typing import Any
+
+import pytest
+from homeassistant.core import State, CoreState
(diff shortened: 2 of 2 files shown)


nielsfaber:
Thank you for taking care of this, it looks like a small but effective fix so I'm fine with merging this.
Regarding:
> On a cold boot, Alarmo can fail to create its alarm entity entirely. The panel then shows "This entity is no longer being provided by the alarmo integration", and the alarm is silently gone until the config entry is reloaded.

There seems to be an underlying issue which is not addressed by this fix.
I think it would be valuable if you find a way to reproduce this issue (I don't experience this).
From there we can look into the root cause of the problem.

PR #1463, opened 2026-08-10
Allow 'Delay on' up to 5 minutes

### LINKED ISSUE #1434: Feature Request: Allow setting 'Delay on' higher than 1 minute
### Checklist
- [x] My request is generic, other users may benefit from it too.

### Proposal
Allow setting 'Delay on' higher than 1 minute so we can use camera person detection to trigger the alarm when loitering is detected. Camera would detect a person in a zone and the sensor should only trigger the alarm after a set loitering time, which we would set in the 'Delay on' field, or a new 'Loitering time' field.
The delay on field currently allows to set a value up to 1 minute, however, we have multiple cases where loitering is only considered after a person is in a zone for more than 3 minutes.

Another use case is for garage gates, we'd like to be able to trigger the alarm only if the gate has been left open for a while. 1 minute isn't sufficient as entering the garage usually takes longer than that.

### Additional info
[some example of how the new functionality should look like]

dennismdejong:
Fixes #1434

Raises the per-sensor **Delay on** picker maximum from 60 s to 300 s (5 minutes), as agreed with the maintainer in the issue.

- Frontend-only change (`sensor-editor-card.ts`): `max="60"` → `max="300"`, step 5.
- Rebuilt bundle committed.
- No backend/storage change — the backend already accepts any positive integer for `delay_on`.

Supports the use cases from the request (camera loitering detection, garage gates) while keeping a security-reasonable cap.

### DIFF
--- a/custom_components/alarmo/frontend/dist/alarm-panel.js
+++ b/custom_components/alarmo/frontend/dist/alarm-panel.js
@@ -2092,7 +2092,7 @@ const qs=2;class Ps{constructor(e){}get _$AU(){return this._$AM._$AU}_$AT(e,a,i)
 
                   <alarmo-duration-picker
                     .hass=${this.hass}
-                    max="60"
+                    max="300"
                     step="5"
                     placeholder="-"
                     ?disabled=${!ms(this.data.delay_on)}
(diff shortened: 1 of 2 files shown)


nielsfaber:
Thank you 👍

PR #1467, opened 2026-08-17
feat: add open_sensors_with_area wildcard to notifications

walnuss0815:
## Motivation

Home Assistant 2026.8 introduced the configurable entity ID generator (<https://www.home-assistant.io/blog/2026/08/05/release-20268/#your-entity-ids-your-choice>). Previously my entities followed the `<type> <area> [<name>]` pattern, so the `{{open_sensors}}` wildcard always made it obvious which room a sensor was in. With the new generator my entity IDs use the `<area>_<device>_<entity>` layout to guarantee uniqueness, and since I removed the area from the device names, several devices now share the same name in different areas (e.g. "Window Left", "Light Ceiling"). The device name alone is therefore ambiguous, and the area of the device is required to identify the sensor.

## Change

Add a new `{{open_sensors_with_area}}` notification wildcard. It behaves like `{{open_sensors}}` but appends the Home Assistant area of each open sensor, e.g. "Left window is open in bathroom".

- The area is resolved through the entity, device and area registries; sensors without an area or without a registry entry fall back to the plain wording.
- The existing device-class based phrasing is reused for the state part, so both wildcard variants produce consistent text (e.g. a motion sensor reads "is detecting motion" in both).
- The `lang` and `format=short` modifiers are accepted in any order; `{{open_sensors_with_area|format=short}}` renders "Left Window (Bathroom)".
- The notification editor offers the new option, and both the backend translations and the frontend localize files were extended.
- The README wildcard table documents the new wildcard.

## Testing

- Added unit tests covering the new wildcard: with an area, without an area, with an empty sensor list, mixed usage of both wildcards, modifier ordering, device without an area, and non-open states.
- Full test suite passes and ruff/eslint/tsc checks are clean.


nielsfaber:
Hello, thank you for this PR.
Is there an issue topic where this feature is discussed? I don't recall any recent discussion about the subject.
I have some problems with this addition.
The main issue is that Alarmo has its own implementation of _areas_. Mixing this with HA's _areas_  makes things very confusing IMO.
I think it would be better to give user the ability of choosing the display setting for sensors. This also keeps things consistent: the sensor name will be displayed everywhere in the same way.
What are your thoughts on this?

walnuss0815:
Thanks for the feedback, and fair point on the issue — I should have opened one before starting on the implementation, sorry about that. Happy to open one now if you'd like the design discussed there instead of on the PR.

I agree with your concern. A persistent, consistent display setting makes sense to me instead of an additional wildcard. But how should the configuration interface look like? I could add a format string to general settings to configure the sensor display name. Possible placeholders could be sensor name, sensor entity id, home assistant area of the sensor, alarmo area of the sensor.

An alternative approach could be to add a simple check box to general settings to add the home assistant area of the sensor to the sensor name.

Once we're aligned on the design, I will rework the PR.

PR #1474, opened 2026-09-03
fix: stable unique_id for Alarmo entities to prevent ID collisions

tabascoz:
The unique_id for alarm_control_panel entities was previously derived from the entity_id (e.g. alarm_control_panel.master_alarm). It was unstable — if Home Assistant renames an entity due to a name collision (appending _2), the unique_id no longer matches, causing duplicate entity errors on restart and the master panel showing only 'disarmed'.

Root cause: AlarmoBaseEntity.unique_id returned self.entity_id, which is based on the human-readable config name, not a stable identifier.

Changes:
- AlarmoBaseEntity.unique_id now raises NotImplementedError, forcing subclasses to provide a stable implementation.
- AlarmoAreaEntity.unique_id uses the stable area_id from the store: f{DOMAIN}_area_{self.area_id}
- AlarmoMasterEntity.unique_id uses a fixed master identifier: f{DOMAIN}_master
- On first load after upgrade, async_add_alarm_entity and async_add_alarm_master migrate the entity registry by updating any entry whose unique_id still uses the old entity_id-based format.
- Orphaned duplicate entities (e.g. master_alarm_2 with the old unique_id) are automatically cleaned up from the registry.

The migration is transparent — existing entity_ids, automations, and history are preserved.

### DIFF
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -132,13 +132,13 @@ async def async_setup_entry(hass, config_entry, async_add_devices):
     @callback
     def async_add_alarm_master(config: dict):
         """Add each entity as Alarm Control Panel."""
+        entity_id = f"{PLATFORM}.{slugify(config['name'])}"
+        new_unique_id = f"{const.DOMAIN}_master"
         _LOGGER.warning(
-            "DEBUG async_add_alarm_master: entity_id=%s hass.data master=%s",
+            "DEBUG async_add_alarm_master called: entity_id=%s master_in_hass=%s",
             entity_id,
             getattr(hass.data[const.DOMAIN].get("master"), "entity_id", None),
         )
-        entity_id = f"{PLATFORM}.{slugify(config['name'])}"
-        new_unique_id = f"{const.DOMAIN}_master"
 
         entity_registry = er.async_get(hass)
         deleted_key = (PLATFORM, const.DOMAIN, new_unique_id)
--- a/custom_components/alarmo/alarm_control_panel.py
+++ b/custom_components/alarmo/alarm_control_panel.py
@@ -41,6 +41,7 @@ from homeassistant.components.alarm_control_panel import (
 )
 
 from . import const
+from homeassistant.helpers import entity_registry as er
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


tabascoz:
After deploying the initial fix, I kept hitting two further issues that took more digging to understand:

**1. async_remove doesn't actually free the entity_id.** When the migration code called `async_remove("master_alarm_2")`, HA moves the entry into a `deleted_entities` cache keyed by `(platform, domain, unique_id)`. On the very next `async_get_or_create()` call (from `async_add_devices`), HA pops that tombstone and *restores* the old `master_alarm_2` name. So every cleanup attempt was being silently undone by HA's own entity registry restoration logic. The fix is to `pop()` the tombstone from `deleted_entities` directly before calling `async_get_or_create`, then rename any live stale entry in place via `async_update_entity(new_entity_id=...)`.

**2. The master panel showing only "disarmed".** This is a startup race. The master's `async_added_to_hass` does two things:
   1. Restores `self._state` from `async_get_last_state()` (typically "disarmed").
   2. Subscribes to `alarmo_state_updated` from areas.

But the area entities often complete their own `async_added_to_hass` and fire their state changes *before* the master subscribes, so those events are lost. The master then runs `async_update_state()` with `state=None` (areas not yet visible), restores "disarmed" from last state, and writes that to HA. When areas finally do fire, `async_update_state()` computes `state="disarmed"` which equals `self._state="disarmed"`, so the early-return guard fires and `update_ready_to_arm_modes()` is never called. Result: frontend shows only "disarmed" with no arm buttons. The fix is to add `self._ready_to_arm_modes` to the early-return condition — if the list is empty, never short-circuit.

**3. Defence against other integrations.** When testing on my own setup I noticed `autoarm` would occasionally call `hass.states.async_set("alarm_control_panel.master_alarm", ...)` directly when its service-call fallback failed. That bypasses Alarmo and leaves a stale entry in HA's state machine that survives across restarts. On the next Alarmo setup, HA sees `master_alarm` as "already in use" in the state machine and renames the new entity to `_2`. I added `_purge_stale_state_entries()` to clear any state-machine entries for our entity_ids immediately before `async_add_devices` runs.

nielsfaber:
Thank you for your efforts on this PR!
A fix for the unique_id issue is very welcome, as it is reported by multiple users and I had no idea how to address this.
Do you consider the PR ready for review/merging?

I think the reported issues 2+3 are not directly related to this PR, so perhaps these should be discussed separately.
For issue 1 however, it seems related and may cause duplicate id's.
What are your thoughts on this?

